Issue #228, opened 2020-05-03
Overhaul Design Recipe templates


schanzer:
*Ignore for now: Not important yet, but it will be*

Our DR layout dates back to 2014, when I first sat down and wrote the DOM specification for this part of Kathi's Scribble build. It's had some known weak spots (*cough* conditionals *cough*), but for the most part it's been pretty reliable for the last six years.

Now we're also trying to build in Pyret, which has a fork of this layout that Dorai made. It works shockingly well for most cases, but it's showing its age:

- [ ] multi-line purpose states don't get multiple comment lines, when they should
- [ ] function names take up a large minimum space, so small names waste a lot of it
- [ ] example/fn bodies that are too long to fit on one line don't expand to take up all of the second line
- [ ] `cond`/`ask` rendering is SUPER brittle
- [ ] The entire system is dependent on `font-face` and `font-size`. If we change fonts or make text larger (accessibility!), the whole thing falls apart.
- [ ] GDrive import is pretty broken

I think we can do a better job here by using `inline-block` and `display:table*` settings, and offloading as much of the line-break behavior to the browser engine / `wkhtmltopdf`. We should make sure to use SPANs for everything that doesn't force a line break, and DIVs for everything that does.

schanzer:
If flexbox winds up being a dead end, we may have to experiment with [Puppeteer](https://github.com/gutschilla/elixir-pdf-generator)

PR #600, opened 2021-09-17
Maximize shared CSS between print and screen

schanzer:
This pull request (PR) includes _more than a thousand changed lines_, almost all of which involve some form of CSS simplification. 

* Replace most (though not all) absolute `px`/`pt` measurements with _relative_ measurements (`rem`), making us less sensitive to font-size changes
* Remove most (though not all) print-specific style rules involving size changes. As of now, the only substantive change is font-size.
* Declare a css `page` type for landscape pages, then use that for a special `@page` rule that handles landscape printing. This removes ALL landscape-specific code from `html2pdf.js`.
* For WB pages with table-based layout (ie - nearly all the circle of evaluation pages), force the table to 90-95% page height. This spaces out the rows evenly, and allows us to get rid of brittle `@vspace` directives.

@flannery-denny and @retabak -- to test this out on a pathway, run the following:
```
git checkout unify-css
./build-pathway -F --book <pathway name>
```

I've gone over every PDF a number of times, but could really use an extra pair of eyes.


retabak:
Feedback on PDF (I looked at OK workbook-long) for @schanzer 

Most persistent issue (which maybe doesn't bother you, but I noted it everytime I saw it): Oftentimes, just one line of text lingers onto a 2nd workbook page (rather than the entire activity fitting into a single workbook page).

- there are no printed page numbers! 
- p.8 & p.15-17 - not enough room for students to draw circles; these should be formatted like p.11 (more space)
- p.35 - dark black lines in upper right-hand section are unnecessary
- p.50 - is "answers may vary" supposed to be printed in the student edition, or the teacher's?
- p.77-78 - is there any way for these two pages to fit onto one page? It's confusing that just one example is on p.78
- p.88 - not a formatting issue, but can we change the def'n of surface area (middle of the page) to read "the sum of the areas of each of its six **faces**" rather than "six sides"?
- p.91 - not a formatting issue - I don't get how this workbook page works but am assuming there are additional instructions somewhere else?
- p.98, ex4 - "string length" gets kinda obscured by the circle
- p.101 - I think the vertically oriented numbers might be hard to read once this is printed / photocopied. Seems like maybe the screenshots of the number lines need to be taken at a higher resolution? Also thinking this page would be a bummer for folks who need to photocopy in black and white. (It is essential for students to see red/green to do the activity.) I'm sure teachers will find work-arounds.
- p.112-113 - The "No" in the sample problem is bolded, italicized and capitalized on p.112 but just bolded and italicized on p.113. 
- p.120-121 - Can these fit onto one page? (Just one question is on p.121.)
- p.126 - not a formatting issue, but: the phrase "starting value" is used in the directions (rather than "y-intercept"). Why?Confusing.
- p.128 - formatting of the sample answer is different from the formatting of the remaining problems. ("y-intercept" should be on its own line.)
- p.131 - maybe not possilbe, but it would be nice to have some white space between graphs. They seem all smushed together. (As a teacher, I'd also put "Pyret notation:" and "function notation:" in the blank box on the right - but that's not essential.)
- p.132 - not a formatting issue - "Function Notation" is capitalized here but not elsewhere.
- p.156 - seems like margins / tabs are off. Three of the examples begin farther left than the others.
- p.193-194 - can these fit onto one page?
- p.197-198 - can these fit onto one page?
- p.206 - can this fit onto p.205?
- p.208 - same issue; can this join p.207?

flannery-denny:
DS workbook. Not seeing page numbers. Not sure why

- [ ] Booleans - can we make the the "Computer Returns" @fitb be the right length now? (pdf p 10)
- [ ] applying functions - table proportioning is weird - should preference space for answer writing (pdf p 11)
- [ ] Lookup questions could use more vertical space between the three sections  (pdf p 26)
- [ ] Chaining methods has gender when it should have sex for #8 (pdf p 44)
- [ ] Adopt formatting from my dataset (pdf p. 55 for the new page I made about can it be answered - it's much clearer)
- [ ] page 51 of pdf. landscape. title box and table aren't the same width. can we get them to match on the left and right sides? Same thing is true with contract pages (probably all horizontal pages)
- [ ] page 62. Is it possible to get the tops and bottoms of the two histograms to line up with each other?
- [ ] Drawing Predictors page 87 & 88 of pdf should be on a single page

Algebra pyret workbook
- [ ] page 8 of pdf order of operation notes - get the programmed expression at the end to appear on a single line.

schanzer:
@retabak @flannery-denny thanks for all the feedback! Keep in mind that this is NOT a general "anything we want to do better in our workbook?" issue, but rather a focused issue dealing with _what breaks_ with all the new CSS. @flannery-denny I've moved many of your comments to a separate issue (#602 ) so that I can address them after the merge.

The goal here is to get this new code into master, after ensuring that the thousand+ changes don't make anything _worse_.

retabak:
@schanzer 

- [ ] p.75-76 are still two pages 
image

- [ ] p.118-119 are still two pages
image

schanzer:
@flannery-denny @retabak I was able to reproduce (and fix!) most of the issues you both reported. There are a small number that I can't reproduce on my own, so I "tightened" some of the spaces on those pages in the hopes that they will fit on a single page on your machines.

When you have a minute, can you do a `git pull` and force-build to check if these issues are resolved? Thanks in advance!

flannery-denny:
I'm looking at algebra-pyret workbook-long-sols and here's what's not fitting on the page:

- [ ] function composition - green star (page 50 & 51)
- [ ] defining values - chinese flag (page 57 & 58)
- [ ] which values would it make sense to define (page 60 & 61)

I'm looking at algebra-wescheme workbook-long and here are the issues I'm seeing:
- [ ] which values would it make sense to define spills onto a second page (page 58 & 59)
- [ ] _Decomposing Flags fits but the flag heights are inconsistent and the tables may have gotten too small for students to write in.  It would probably be better to shrink flags and let the table cells be slightly taller._
- [ ] matching contracts and examples (page 83) - line break in code for number 5 splits strangely
- [ ] matching contracts and examples 2 (page 84) -  several lines of code split strangely
- [ ] word-problem update-player(page 120) and word-problem: line-length(page 122)... seems these last parentheses probably belong at the right side of the blank line they're on?
Screen Shot 2021-09-23 at 12 16 56 AM

I'm looking at data-science workbook-long and i'm seeing the following issues:
- [ ] triangle contracts sas-asa (page 21) - split in code shading
Screen Shot 2021-09-23 at 12 22 43 AM
] 
- [ ] star-polygon page 22 - same shading split


I'm looking at data-science workbook-long-sols and I'm seeing the following issues:
- [ ] Identifying Form Direction and Strength (matching) page 98 spills onto page 99
- [ ] Drawing predictors page 101 spills onto page 102
- [ ] Interpreting regression lines and r-values ([age 103) is rotate 90 degrees counter clockwise

schanzer:
@flannery-denny thanks - most of these I was able to reproduce and (I believe) fix. 

* I defensively made the function-composition pages a little shorter than they need to be on my machine, to give us some margin of error.
* Something weird is going on with which-value-to-define: sometimes the CSS rules are honored, and sometimes they're not! Trying to track down that bug now.
* Line breaks on 83 and 84 are weird on `master`, too. I made a small change which I _think_ is an improvement, but ultimately there's no fix here unless we shorten the actual example code in the file.
* The parens at the end of update player are in the same place on `master` -- and can't be fixed until #228 lands

Let me know if any of the other issues are still showing up for you?

flannery-denny:
Algebra - WeScheme Long
- [ ] p9 - look at the first circle and notice how close the 4 and the 2 are.  There is similar spacing for other circles on the page. It would be clear if the numbers were twice as far apart.
- [ ] p12 - The @fitb don't need to be so long for COE. And if they weren't so long, the numbers could be spaced more clearly.
- [ ] p23 - evaluating COE - for some reason 6 & 7 are identical (I know this is likely not a css issue)
- [ ] p57 - Which values would it make sense to define is still spilling over. Can't you make all 6 of the flags shorter & the same height? St. vincent and the grenadines is much taller than the Philippines...
- [ ] Not seeing any additional issues in the solutions version

Algebra-pyret long-sols
- [ ] p8 order of operations notes - code splits onto two lines and shouldn't
- [ ] p21 converting coe to arith - number 4 is wrong

Not seeing any problems in DS-long

In DS-long-sol I see:

- [ ] p53 Chaining Order Matter 2 - On most of our pages the number and match are on one line. On this page the solution is beneath the number and it looks less tidy. There is plenty of space.
- [ ] p88 - Matching solutions to histograms. C reads as copyright symbol and for some reason 4 and A are not near each other on the page
- [ ] p105 - Describing Relationships - the answer text needs to be smaller font in order to fit on the number of lines provided.
- [ ] p118-120 - Blank Design Recipes - Any way to give a little more space for writing directions at the top instead of having so much blank space at the bottom?

*None of these issues would break the build. Once you've addressed the ones you can, I think it's fine to merge.*

schanzer:
@flannery-denny I fixed a bunch of the items you raised, but moved them to #602 since they aren't part of this pull request. I also think I've _finally_ snagged the bug that was causing which-value-to-define to overflow.

In theory, this should be ready to merge!

PR #612, opened 2021-09-29
attempt at reorganizing DS teacher resource page

retabak:
Here's an attempt at reorganizing the DS teacher resource page!

The branch name is ds-teacher-resource 

Some notes:
- The first few bullets include redundancies because the asciidoc automatically included links to "Teacher Workbook" and "PD Workbook". (I propose that we refer to these as "Teacher Edition" and "Teacher Answer Key" instead; those names automatically carry distinct meanings to teachers, whereas the terms we are using do not, and are - in my opinion - a bit confusing.) I think Dorai might need to do some work to fix these links so that they are automated?

I also removed the PD pre/post surveys as I do not think they belong on this page. I know that @flannery-denny agrees with this; @schanzer , if it's important to you that they stay, let me know. 

If Dorai develops a new project directive, I would use that too.

### DIFF
--- a/pathways/data-science/langs/en-us/resources/index.adoc
+++ b/pathways/data-science/langs/en-us/resources/index.adoc
@@ -1,76 +1,66 @@
-= Teacher Resources
+== Teacher Resources
 
-++++
-<style>
-.sect1 { border-top: 1px solid #efefed; }
-</style>
-++++
-
-- *Exercises and Solutions* - <<Exercises and Solutions, Jump to Exercises and Solutions>>
-
-- *Starter Files* - <<Starter Files, Jump to Starter Files>>
-
-- *Projects* - <<Projects, Jump to Projects>>
-
-- *Contracts* [@link{solution-pages/contracts.pdf, file}] - a reference PDF of just the contracts pages from the back of the student workbook.
-
-- *Implementation Options* - looking to integrate Bootstrap:Data Science into your classroom? There are @link{pages/implementation-options.adoc, implementation guides} for teachers looking to fill a week's worth of time, or a full-year course!
-
-- *Online Community (Discourse)* [@link{https://discourse.bootstrapworld.org/, link}] - Want to ask a question or pose a lesson idea for other Bootstrap teachers? Want to see resources other teachers are making and sharing? Our discussion forum is the place to do it.
-
-- *Workshop Files* - <<Workshop Files, Jump to Workshop Files>>
-
-- *Collection of Bootstrap:Data Science Desmos Activities* [@link{https://teacher.desmos.com/collection/5fc9250afcaa3f30f4946746, link}] - Love Desmos as much as we do? We have plenty of Desmos activities infused throughout our Data Science lessons, and you can explore all of them in one place.
-
-- *What's Going on in this Graph?* [@link{https://www.nytimes.com/column/whats-going-on-in-this-graph, nytimes.com}] - Once a week, the NYTimes posts an interesting data visualization. This is a great resource to spark the curiosity of young data scientists, and make for a perfect Notice &amp; Wonder activity!
-
-- *The Pudding* [@link{https://pudding.cool/, link}] - Terrific deep dives into data analysis for interesting topics. Another wonderful resource for students, with great animations showing the transformation of how data is visualized as the research questions evolve.
-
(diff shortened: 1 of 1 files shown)


flannery-denny:
Nice work @retabak. Looks much better. I think the contracts pdf belongs up at the top. It's the file I referred to the most when I was in the classroom.

I'm not sold on your new names for the workbook files.
Have you looked at them? I'm pretty sure the PD workbook includes lesson plans and empty activity pages. And the teacher workbook is the answer key. That doesn't mean I think the names they had need to stay, but the names you picked don't make me expect those things. Happy to discuss.

schanzer:
@retabak great first draft!

- The PD Workbook is literally the book we'd like to give people during the PD. But you know what? We haven't used the lessons+student book in years, and nobody seems to have missed it. Perhaps we should remove it....
- Is "teacher answer key" more recognizable than "Workbook with Answers?"
- Why does "animals dataset" belong in starter files?
- Not your fault, but seeing it here I really dislike "Animals Dataset - Part 2". We use it for random sampling, and it has a table called `big-animals-table`. Can we rename the link to something like "Big Animals Table", and add "(used in Randomness and Sample Size)"?
- Having "Other Facilitation Resources" at the bottom makes them almost undiscoverable. Why not include these at the top? Contracts and Implementation Options seem seem especially important to make discoverable

flannery-denny:
Good catch @schanzer. We actually already fixed the name of that animals starter file in the lesson plan... it's called `Expanded Animals Starter File` ... and it's on the list of what we need to fix inside the starter file and workbook pages before the 2022 release... more evidence that autogenerating the starter files list would be a good move!

All of the other datasets are ones that students get to choose from. The two animals dataset starter files are files that teachers need quick access to in order to teach our lessons. I think they should remain with the starter files list.

What do you think @retabak?

retabak:
Re: putting the datasets with the starter files---doesn't really matter to me.  I think the datasets show up when they are needed in the various lessons, _and_ I don't think this page is really where folks go for "quick access."  TBH... if it were up to me, we wouldn't put the starter files on this page at all. The page is *very* long already, and I honestly can't think of a scenario where a teacher would be looking for JUST a starter file.  Those starter files are so integrated into the lessons they were designed for.

(As I type this, I recognize that the teacher resource page is an index of sorts. I am ready to lose this battle!)

As for the titles of those workbooks... I have a lot of feelings, many of which are informed by the fact that I did not learn about BS through a PD, nor did I attend a PD for years. So:
- I believe that a "Teacher Edition" is exactly what you, @flannery-denny, have described: lesson plans and blank student pages. What do you think of when you hear "Teacher Edition"? What makes this a "PD Workbook"? _Is it only something one would use at a BS PD?_ What exactly _is_ a "PD Workbook" anyhow? This PDF seems like a guide that I could use to facilitate instruction in my classroom: I can see all the lesson narratives, and I can see what my students have access to also--it's all in one place.
- On that note... **mentioning PD in passing will, I am almost certain, confuse teachers who have _not_ attended PD and make them wonder if perhaps they can't teach our curriculum unless they go to PD**.  I don't think we want to do this! We advertise that our curriculum is free, that we provide EVERYTHING on our website--but then there are these few little passing mentions of PD peppered throughout our webpage. It feels like mixed messaging. 
- @schanzer, we don't say "Workbook with Answers," we say "Teacher Workbook."  "Teacher Answer Key" is widely used by other curricula, but I suppose "Student Workbook with Answers" would suffice also. "Teacher Workbook," however, makes me think that this is something that it most definitely is not: a workbook for teachers to complete--like to train them how to code or to train them how to teach.
- I realized that the "PD Workbook" on the DS resource page just links to a student workbook. So, woops! We should fix that.

flannery-denny:
I think Teacher Answer Key is very clear - let's change the name.
I vote we get rid of the PD book. Our lesson plans make most sense in their
dynamic webpage format and we're constantly improving them. They can be
printed individually. I don't think we need to encourage folks to print
them en masse.

retabak:
Cool! Sounds like everyone is OK with getting rid of the PD book!

schanzer:
@ds26gte can you comment out the code that generates the PD workbook? We may want it back in the future, but no reason to generate it now.

@retabak for the workbook names, how about we have "Workbook" and "Workbook (Teacher's Edition)". When I was teaching, the teacher's edition of the textbook always came with the answers. And since all our workbooks are for students, there's no need to call it "Student Workbook"

retabak:
@schanzer - let's go with renaming the PDFs "Workbook" and "Workbook (Teacher's Edition)" ! Thanks.

schanzer:
@retabak I've just pushed the change. If it looks good, I can merge the PR!

retabak:
@schanzer - awesome! I want to go back and make sure I've addressed all feedback in this issue. Hoping to do so by tonight or tomorrow at the latest.

PR #617, opened 2021-10-05
Smarter fill-in-the-blanks

schanzer:
This PR introduces `display:flex` to our repetoire for styling fill in the blanks. For now, its use is limited to Design Recipes only, as it addresses #228 . Immediately, this results in all blanks growing to fit the available horizontal space.

Over time, we can expand this use into other worksheets, replacing brittle `@fitb{XYZ}` with `@fitb{flex}` to show that a blank should extend as far as possible. This makes maintenance a lot easier in the long run, since blanks "know how to stretch themselves".

@flannery-denny and @retabak , you can review this by checking out the `better-dr` branch, and building the `__sample` pathway. I've pre-populated it with virtually every lesson that uses the Design Recipe, so if those look good to you we should be in good shape. None of these changes should effect non-DR pages, but I did a quick eyeball of the Algebra and Data Science pathways to make sure. Nothing jumped out at me, but if you have time it wouldn't hurt to get quick check on those as well.

**Please only comment  on regressions from master** - if there's something you dislike about a page, make sure it's truly unique to this branch.

### DIFF
--- a/lib/shared.css
+++ b/lib/shared.css
@@ -485,7 +485,7 @@ table.recipe_title td { padding: 0px; margin: 0px; border: none; }
 /* Note(Emmanuel): these could probably be a lot tighter */
 .studentAnswerCode, .studentAnswer, .studentAnswerUnfilled, 
 .studentAnswerBlockUnfilled .studentBlockAnswerUnfilled,
-.studentBlockAnswerFilled, .pyretkeyword {
+.studentBlockAnswerFilled {
   margin-bottom:      0.85rem;
   min-width:          105pt;
   position:           relative;
@@ -494,7 +494,8 @@ table.recipe_title td { padding: 0px; margin: 0px; border: none; }
   display:            inline-block;
   border-bottom:      solid 1px black;
 }
-.studentAnswerCode::after, .studentAnswer::after, .pyretkeyword::after {
+.pyretkeyword { margin: 0.25rem 0; }
+.studentAnswerCode::after, .studentAnswer::after {
   position:           absolute;
   bottom:             -1rem;
   left:               0;


schanzer:
@flannery-denny can you define "shrumpfed"? Are letters piling onto one another? Is the spacing between words missing? I don't see any of that on my end, so it's hard for me to understand what you mean.

The other items you describe ask for a minimum size for blanks, which makes sense (there actually _is_ one, but clearly it needs to be larger!). But you've described the contract stuff differently, which makes me think you're seeing something else here.

I'll set minimum widths and push an update later this morning.

flannery-denny:
shrumpfed - got smaller.  We've got a whole other inch on the contract line. It used to take up the full width. I think there are lots of function-names that students won't be able to write in the allotted  space and we should give them as much space as possible here.
Screen Shot 2021-10-06 at 8 51 02 AM

schanzer:
That's how it used to be. Here's why we changed it:

In Pyret, one of the most common errors is people inserting whitespace between the function name and the parentheses around the variables. There seems to be some predisposition people have towards putting a space here, which asserts  when the solutions have even a _little_ bit of space. Moving the function name to the right seems to have reduced this somewhat.

Does that change your thinking about this?

flannery-denny:
Heard.

If the size of the line is flexible, it would read better if the lines that have text sitting on top of them were shorter, so the words could be on the left of the line... and next to the colon as you say is important.

If that would take a lot of investment, we can try leaving it as is, since you think it works better. 

Even if it doesn't look as good, it is an architectural rule that form should follow function.

schanzer:
That was my original thinking, which is why those lines were so short! The problem is that kids _always_ need more room to write each character than a computer does. So if I remove the min-width and use only _exactly_ as much space as we need (your current proposal), we'll wind up with not enough room for kids to write (your previous proposal). I'm willing to spend another hour seeing if I can get it to generate one way for solutions and another way for students. Is that worth it to you?

flannery-denny:
Yes. I think it will make a big difference in how the pages read and there are lots of dr pages. 

But, if it’s too much work, space for kids to write has to trump solutions.

Flannery

flannery-denny:
@schanzer. I want to say yes...
But...
the function name is landing on top of the word function name.

[design-recipe-compute.pdf](https://github.com/bootstrapworld/curriculum/files/7304919/design-recipe-compute.pdf)

PR #1172, opened 2022-10-18
Standards audit (closes #968)

### LINKED ISSUE #968: Audit of standards alignments
It appears that only two (!!) of our lessons are aligned to K12CS standards - defining-functions and defining-values.
https://bootstrapworld.org/materials/latest/en-us/courses/algebra-pyret/pathway-alignments.shtml

I discovered this when searching out a standard to assign to "Computing Needs All Voices." (I found one! In the K12CS standards!) I think we ought to go through and assign K12CS standards to each lesson. (Unless there's a reason it was never done?) 

I also noticed that the Iowa alignment appears to be incomplete. (Only four of our lessons are aligned to these standards.) I also see that we have a .rkt doc of NGSS standards, but they have never been assigned to lessons either. So it looks like standards alignment in general would benefit from a little TLC.

schanzer:


schanzer:
@ds26gte when I build a/any lesson plan, it looks like some of the state standards are correctly hidden-until-selected, but many are not. From the web inspector, it looks like some are being generated with improper class names (see below). Can you investigate when you get a chance?

image

PR #1176, opened 2022-10-21
Better dr

### LINKED ISSUE #1151: Rework DR template with more use of CSS flex
As I learn more about how `display:flex` works and what it's capable of, I'm thinking we can make our DR generation simpler, more maintainable, and less brittle. The trick will be teaching the browser how to reflow lines properly when they wrap, but it's increasingly looking like I could apply the same trick from #840 to this.

- [x] use `fitruby` dom/CSS instead of custom `.studentAnswer`, if possible. 
- [x] use flex for cond and if/then clauses, wrapping each one

schanzer:
Remove the `.studentAnswer` class, using `.fitbruby` instead. Closes #1151 

Remaining todos
- [x] cleanup rubies, and try to preserve the i18n-via-css mechanism we've been using thus far
- [x] see if we can't also eliminate `studentAnswerCode`, `studentAnswerFilled` and `studentAnswerUnfilled`
- [x] Remove legacy support in the preprocessor


schanzer:
@ds26gte I think we're ready for the next step, which is to replace `studentAnswerFilled` and `studentAnswerUnfilled` with just `studentAnswer`, and stick the `solution` class on the body element

PR #1186, opened 2022-11-02
Early math

retabak:
Right now, there are 12 very rough drafts of lessons.

Worksheets are largely missing, but - if you'd like - you can (sometimes) get a sense of what I am envisioning by checking my sketches in this [GDrive folder](https://drive.google.com/drive/folders/1ncImyvK5utQ5RL2tH_bmOfYY-Q5JZKiV?usp=share_link). A few of the sketches only live in my brain, unfortunately.

That said, you probably don't need to go digging for worksheets because this is a SUPER preliminary review, during which I hope you will consider the following:

- are the circles advancing the learning objectives? or does this just seem like a convoluted way of teaching early math?
- does the sequencing of lessons make sense? I'd like there to be a coherent narrative arc but I don't think I've achieved that.

Those two things are the most important, at the moment. One final, less important consideration...
- As you're reading, if you think of any ideas on how to spice up the lessons (i.e. role play! games! projects! things-that-aren't-worksheets!), I'd love to hear about them and weave some of them in. As I already articulated to @schanzer, getting on computers and coding breaks up the onslaught of Algebra / DS worksheets. That doesn't happen in this pathway. (This seems like something I will likely address a bit later on.)

PS - Hoping I did OK making my first PR. Open to feedback there, too, in case I messed up.
PPS - It's actually my 2nd PR. Closed the first and reposted at @schanzer's request.


lessons/equivalence-and-laws/langs/en-us/index.adoc
```diff
@@ -0,0 +1,167 @@
...
+// https://illustrativemathematics.blog/2018/04/10/what-is-an-instructional-routine/ //
+// http://www.meaningfulmathmoments.com/which-one-doesnt-belong-wodb.html //
+
+
+
+
+=== Synthesize
+
+- Are equations with variables always true?
+** _No. If we're looking at an equation that represents one of our Circle Transformations, then yes. An equation such as @math{j - y = y - j} is only sometimes true. And an equation like @math{g \times 0 = 600} is never true._
```
schanzer, line 165: I think you're still learning too hard on "Circle" as a vocabulary word. I get that you're trying to show how cool Circles are, but I think this actually works against that goal. The circles are just a way of highlighting structure, and we want teachers to see them as a small cost with a big benefit. Inventing vocabulary like "Circle Transformations" makes it seem like a larger cost: should teachers quiz kids on what Circles Transformations are? And since Circles aren't in the standards, this feels like a bigger cost with zero benefit.

We're looking at equations that differ according to properties of Commutativity and Associativity. When they are validly applied to get from one equation to another, we know they are equivalent. When they are _not_ applied validly, we don't know if they are equivalent.  You can make the same point without adding vocabulary

lessons/identity-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,109 @@
+= The Identity Property
+
+@description{This lesson introduces students to the Commutative Property by way of Circle Transformations.}
```
schanzer, line 3: ?

lessons/identity-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,109 @@
...
+|[.materials-links]
+@material-links
+
+|===
+
+== The Identity Property
+
+=== Overview
+
+Students analyze pairs of Circles to see if they represent equivalent expressions. The Circles make visible the structural changes that occur when we apply the Commutative Property.
```
schanzer, line 36: ?

lessons/identity-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+=== Launch
+
+Build on prior knowledge by reminding students of the activity that they completed in the previous lesson. You might even refer to the list of category names that the class created, and then draw students’ attention to any category names that allude to or suggest the commutativity. (Any mention of “changing the order” is a strong hint that students’ have clued into the Commutative Property!)
+
+[.lesson-instruction]
+- Now, turn to Discover the Identity Property. Look at the first problem. Translate the expressions in words into Circles of Evaluation. What do you Notice? What do you Wonder?
```
schanzer, line 43: Hard for me to comment on this without seeing the worksheet

lessons/identity-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,109 @@
...
+|@show{(coe '(+ 3 q))}	| @math{=} | @show{(coe '(* (+ 3 q) 1))}
+| @math{3 + q} 	| @math{=} | @math{(3 + q) \times 1}
+|===
+
+Or like this:
+
+[.embedded, cols="^.^3,^.^1,^.^3", grid="none", stripes="none" frame="none"]
+|===
+|@show{(coe '(+ 6 t))}	| @math{=} | @show{(coe '(+ 0 (+ 6 t)))}
+| @math{6 + t} 			| @math{=} | @math{(6 + t) + 0}
```
schanzer, line 87: Why use different numbers and letters here? I would use q and 3 in both, to really show how it's the associative property and the expressions are equivalent

lessons/identity-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,109 @@
...
+[.lesson-instruction]
+- Complete TRUE or FALSE with Variables using your knowledge of the Identity Property.
+- Analyze the Circles to decide Which One Doesn’t Belong? with Variables. Be sure to explain your thinking.
+- Complete Label the Arrow with Variables. Identify if each Circle Transformation represents the Identity Property, the Commutative Property, the Associative Property, or Neither.
+
+=== Synthesize
+
+- Did you use Computation to check your work? Or do you prefer thinking about Circle Transformations and equivalence?
+** _Student responses will vary._
+- There is a version of the Identity Property for each of the four operations - addition, subtraction, multiplication, and division. This is *not* the case for the Commutative Property or the Associative Property. Why is this so? How is the Identity Property different from these other properties?
```
schanzer, line 107: Love this!

lessons/absolute-value-and-opposites/langs/en-us/index.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+[.lesson-instruction]
+- What number is the same distance from zero as @math{8}?
+** _@math{-8}_
+- What number is the same distance from zero as @math{-3}?
+** _@math{3}_
+- The pairs of numbers we've just identified are called _opposites_ - two numbers that are the same distance from zero on the number line, with one negative and one positive.
+- What are some other pairs of opposites that you can identify?
+** _Possible responses: @math{6.5} and @math{-6.5}; @math{4} and @math{-4}; @math{\frac {7}{3}} and @math{- \frac {7}{3}}._
+- Do you think that zero has an opposite?
```
schanzer, line 57: How about starting with a trickier question: One number on this line doesn't have an opposite. What is it? 

Or perhaps even better: are there any numbers on this line that don't have an opposite?

lessons/absolute-value-and-opposites/langs/en-us/index.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+We annotate absolute value like this: @math{|x|}, with @math{x} being any given number. When we encounter an expression like @math{|x|}, we say "the absolute value of @math{x}."
+
+Because _opposites_ are the same distance away from zero, they will always have the same absolute value. So, @math{|4| = 4} and @math{|-4| = 4}.
+
+// include number line diagram showing a 4-unit jump from 0 to 4, and from 0 to -4 //
+
+=== Investigate
+
+The function to represent absolute value is `num-abs`. It can be used like any other operator that we put at the top of a Circle of Evaluation - with one important difference: when we use `negate`, we include just one argument inside the Circle.
```
schanzer, line 117: Given that you already asked to make `negate` a function, is it worth also defining `abs`?

lessons/evaluating-exponents/langs/en-us/index.adoc
```diff
@@ -0,0 +1,116 @@
...
+[.lesson-instruction]
+- Show students the following expression: @math{2^3}
+- One way of reading this expression is "two raised to the power of three." Do you know any other ways of reading this expression with an exponent?
+** _Two to the third power; two to the third; two cubed._
+- What does it mean to _cube_ a number?
+** We multiply it by itself three times.
+- What does this expression evaluate to?
+** @math{2^3 = 2 \times 2 \times 2 = 8}
+
+Clarify and reinforce some key concepts about exponents via class discussion:
```
schanzer, line 52: I think this misses an opportunity to justify why exponentiation is valuable: _it's a shorthand_. Show the Circle of Eval for 2^3 (or a higher power!), and point out how using exponents makes this much cleaner.

`num-expt(2,8)` is a much nicer way of writing 2 * 2 * 2 * 2 * 2 * 2 * 2 * 2

lessons/evaluating-exponents/langs/en-us/index.adoc
```diff
@@ -0,0 +1,116 @@
...
+Now, have students analyze the two Circles below.
+
+[.embedded, cols="^.^1,^.^1", grid="none", stripes="none" frame="none"]
+|===
+|@math{5^2}							| @math{2^5}
+|@show{(coe  '(num-expt 5 2))}		| @show{(coe  '(num-expt 2 5))}
+|===
+
+[.lesson-instruction]
+- Can we switch the order of the exponent and the base, as we could when adding or multiplying values?
```
schanzer, line 77: You spent all this time building up the term "equivalence", finally introduced it a few lessons ago, and now it's gone! Let's _use it_:  "is it TRUE or FALSE that the following two expressions are equivalent? Why?"

schanzer (changes requested):
@retabak this is a terrific start! I left a few comments throughout the PR, ranging from high-level to low-level. In a few places I couldn't really chime in because without a wbpage I couldn't "see" that part of the lesson. I'll come back to these when the wbpages are a little more fleshed out.

lessons/associative-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+== The Associative Property @duration{30 minutes}
+
+=== Overview
+
+Students analyze pairs of Circles to determine if they have been manipulated in a valid way. The Circles make visible the structural changes that occur when we apply the Associative Property.
+
+=== Launch
+
+The Associative Property defines an equivalence relation over structures, such that "adding 4 to the sum of 3 and 2" is equivalent to "adding the sum of 4 and 3 to 2." In the short term, familiarity with the Associative Property enables students to make mental computations more efficiently. In the long term, comfort with the laws of arithmetic will provide a strong foundation for more complex algebraic reasoning.
```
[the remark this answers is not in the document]
retabak: Lol! I had a crappy definition of the associative property in an early draft which you critiqued as syntactic (it was).... so I literally stole a sentence that you wrote. I can't figure out how to write a good explanation of this property! I'll play with it some more.

lessons/absolute-value-and-opposites-more/langs/en-us/index.adoc
```diff
@@ -0,0 +1,143 @@
...
+
+Circulate as students sort their cards, ensuring that they are carefully analyzing each card so that they see the connection between the Circle of Evaluation and its translation into words. Students will need to consider if translating the Circle into words involves reading from the outside-in or the inside-out.
+
+[.lesson-instruction]
+- Now, you’re going to receive a set of C cards.
+- One at a time, take turns analyzing each C card. After examining it closely, place it with the appropriate Circle and expression in words.
+- Explain to your partner how and why you placed each card.
+- You and your partner must agree on each card’s placement before advancing to the next.
+
+Again, ciculate during the activity. As students match expressions in words to their arithmetic expressions and Circles, invite them to use the Circle as a tool to help them translate to the other representations.
```
flannery-denny, line 63: ciRculate

lessons/translating-words-math/langs/en-us/index.adoc
```diff
@@ -0,0 +1,176 @@
...
+Share these two key concepts before explaining the following instructions.
+
+[.lesson-instruction]
+- With your partner, match each A card with the corresponding B card.
+- Lay them out on the table in front of you so you can clearly see both the Circle and the expression. Create a separate pile for any cards that do not have a match.
+- Discuss each match that you make.
+
+Circulate as students sort their cards, ensuring that they are carefully analyzing each card so that they see the connection between the arithmetic expression and the Circle of evaluation. This first phase of matching cards should be relatively quick - but it might raise questions on a few topics:
+
+- *Symbols*. Circles utilize * to represent multiplication and / to represent division. (Why? Circles are a bridge representation - one which can eventually be used to help students learn to code! Although we use these symbols on all of our materials, you and your students can use whichever operator symbols are most comfortable.)
```
flannery-denny, line 65: recommend using some special formatting to make `*` and `/` stand out. I had to read this a few times to make sense of it.

lessons/translating-words-math/langs/en-us/index.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+- *Position of the operator*. The operator always belongs at the top of the Circle, and not in between terms.
+
+[.lesson-instruction]
+- Now, you’re going to receive a set of C cards.
+- One at a time, take turns reading each C card out loud. After reading the card, place it with the appropriate Circle and expression.
+- Explain to your partner how and why you placed each card.
+- You and your partner must agree on each card’s placement before advancing to the next.
+
+Again, ciculate during the activity. As students match expressions in words to their arithmetic expressions and Circles, additional challenges may arise.
```
flannery-denny, line 77: ciRculate

lessons/translating-words-math/langs/en-us/index.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+- *Symbols*. Circles utilize * to represent multiplication and / to represent division. (Why? Circles are a bridge representation - one which can eventually be used to help students learn to code! Although we use these symbols on all of our materials, you and your students can use whichever operator symbols are most comfortable.)
+
+- *Order of terms*. Although 5 + 3 and 3 + 5 both evaluate to 8, the Circle of Evaluation that corresponds with each of these expressions is unique.
+
+- *Position of the operator*. The operator always belongs at the top of the Circle, and not in between terms.
+
+[.lesson-instruction]
+- Now, you’re going to receive a set of C cards.
+- One at a time, take turns reading each C card out loud. After reading the card, place it with the appropriate Circle and expression.
```
flannery-denny, line 73: As the teacher, I want to know what range of things students might see on the C cards!

lessons/translating-words-math/langs/en-us/index.adoc
```diff
@@ -0,0 +1,176 @@
...
+- Explain to your partner how and why you placed each card.
+- You and your partner must agree on each card’s placement before advancing to the next.
+
+Again, ciculate during the activity. As students match expressions in words to their arithmetic expressions and Circles, additional challenges may arise.
+
+[.lesson-instruction]
+Practice what you’ve learned on @printable-exercise{practice-translating.adoc} and Matching Words to Circles.
```
flannery-denny, line 80: You're going to need to figure out how to leave enough space in this table so that students can actually draw their circles.

lessons/translating-words-math/langs/en-us/index.adoc
```diff
@@ -0,0 +1,176 @@
...
+|===
+
+== Circles of Evaluation @duration{30 minutes}
+
+=== Overview
+Students match Circles of Evaluation to mathematical expressions, and then they consider how mathematical expressions in words map onto Circles of Evaluation.
+
+=== Launch
+
+Explain to students that they are going to receive @handout{pages/three-sets-cards.adoc, "three sets of cards"}.
```
flannery-denny, line 46: curious where this card set is headed. currently there are circles that repeat and blank spots...

lessons/translating-words-math/langs/en-us/index.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+The first set of cards includes mathematical expressions. Students have definitely encountered those before! The expressions on their cards each include the number 15, the number 3, and an operator (+, -, *, or /).
+
+The second set of cards includes _Circles of Evaluation_. Students have almost definitely never encountered Circles of Evaluation before! Broadly, Circles of Evaluation will help students to visualize the structure of the mathematical expressions they encounter. For the time being, here’s what students need to know about Circles of Evaluation:
+
+- Every Circle must have one - and only one! - function, written at the top.
```
flannery-denny, line 52: As the student, I would need to be told from the beginning that circles will be used to model much more complex expressions with nested circles in order to get me to buy into them.

lessons/computation-transformations/langs/en-us/index.adoc
```diff
@@ -0,0 +1,151 @@
+= Computation as a Circle Transformation
```
flannery-denny, line 1: Not sure what a Circle Transformation is or what this title is supposed to mean?

lessons/computation-transformations/langs/en-us/index.adoc
```diff
@@ -0,0 +1,151 @@
...
+== Computation as a Circle Transformation @duration{30 minutes}
+
+=== Overview
+
+Students transform a simple arithmetic expression to create as many equivalent Circles of Evaluation as they can. They then categorize these Circles to consider different possible transformations and to identify _Computation_ as a Circle Transformation.
+
+
+=== Launch
+
+Invite students to take a look at this Circle of Evaluation:
```
flannery-denny, line 48: Curious about this explanation of equivalence and the subsequent request to have students make as many equivalent circles as they can. Not convinced about use of the word equivalence. But, assuming you land on wording that sits well with @schanzer, I would suggest that this activity would make more sense if you started with an expression rather than a circle, then showed students a set of circles that could be used to model the expression and asked them to discuss why each of them worked as a model for the expression, and then gave students an expression which you wanted them to model with as many circles as possible.

lessons/computation-transformations/langs/en-us/index.adoc
```diff
@@ -0,0 +1,151 @@
...
+Now take a look at this example:
+
+
+[.embedded, cols="^.^3,^.^1,^.^3,^.^1,^.^3", grid="none", stripes="none" frame="none"]
+|===
+| @show{(coe '(+ (- 10 8) (* 3 6)))} | &rarr; | @show{(coe '(+ 2 18))} | &rarr; | @math{20}
+|===
+
+- Does the order in which we transform the two inner circles (above) matter?  Why or why not?
+** _No, the order does not matter! We could solve the circle on the left first, or the circle on the right._
```
flannery-denny, line 136: I'd add "The circles are INDEPENDENT of each other."

lessons/equivalence-and-laws/langs/en-us/index.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+@keywords{variable}
+
+[@lesson-intro-table]
+|===
+
+| Lesson Goals
+| Students will be able to...
+
+* Determine if two expressions are equivalent by applying Circle Transformations.
```
flannery-denny, line 15: instead of "by applying Circle Transformations" how about "using Circles of Evaluation"

lessons/commutative-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,116 @@
+= The Commutative Property
+
+@description{This lesson introduces students to the Commutative Property by way of Circle Transformations.}
```
flannery-denny, line 3: Instead of "by way of Circle Transformations" how about "using Circles of Evaluation"

lessons/commutative-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+=== Investigate
+
+[.lesson-instruction]
+- Turn to Commutative Property Table.
+- How many different ways can you apply the Commutative Property? Draw as many different Circles of Evaluation as you can.
+
+This activity is similar to one students’ completed during the Computation as a Circle Transformation lesson.  In both activities, students are developing equivalent expressions. That said, in this lesson, students cannot write down any equivalent circle; they must develop Circle Transformations that apply the Commutative Property.
+
+Note that each of these Circles can be represented in at least one alternative way. Some circles, however, include subtraction and division. You may need to draw some students' attention to the operation as they rearrange the values inside the circle, helping them to notice that the Commutative Property does not apply for division and subtraction.
```
flannery-denny, line 56: An extension of this could involve negative numbers.
Whereas 5 - 6 does not equal 6 - 5, 5 - 6 is equivalent to 5 + -6 which does equal -6 + 5. And this is sixth grade work.

Similarly, for division it might be worth making the connection to fractions.
8 / 2 does not equal 2/ 8 but 8 /2 can be written as 8 * 1/2, which is equivalent to 1/2 * 8

lessons/associative-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+[.embedded, cols="^.^3,^.^1,^.^3", grid="none", stripes="none" frame="none"]
+|===
+
+|@show{(coe '(* (* 2.5 4) 36))}	| &rarr; | @show{(coe  '(* 10 36)) }
+| First, she grouped together the first two factors in their own Circle. |  | Then, she applied Computation as a Circle Transformation.
+
+|===
+
+She then applied Computation as a Circle Transformation. Alecia knows that @math{2.5 \times 4} is @math{10}. She also is confident multiplying by powers of 10. Circles helped her to literally see that the final product is @math{3,642}.
```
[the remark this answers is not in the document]
flannery-denny: this suggestion would be more powerful if the numbers were ordered differently to start. 2.5 x 36 x 4, where someone had to notice that it would simplify things by starting with multiplying 2.5 by 4, rather than that being the default left to right order.

lessons/equivalence-and-laws/langs/en-us/index.adoc
```diff
@@ -0,0 +1,167 @@
...
+|@show{(coe '(* 2 (* 3 c)))}	| @math{=} | @show{(coe '(* (* 2 3) c))}
+| @math{2 \times (3 \times c)} 	| @math{=} | @math{(2 \times 3) \times c)}
+|===
+
+Or like this:
+
+[.embedded, cols="^.^3,^.^1,^.^3", grid="none", stripes="none" frame="none"]
+|===
+|@show{(coe '(+ 5 (+ 6 m)))}	| @math{=} | @show{(coe '(+ (+ 5 6) m))}
+| @math{5 + (2 + m)} 			| @math{=} | @math{(5 + 2) + m}
```
flannery-denny, line 119: noting that the numbers in these expressions do not match the circles.

flannery-denny:
@retabak I'm not an expert in pull requests yet and in the process of leaving specific comments seem to have lost my ability to leave a proper review. So, in addition to my line-specific comments, here's my overarching feedback.

- It's cool to see circles of evaluation being used like this... and yes, they are definitely serving a purpose.
- I 100% agree with @schanzer that inventing language like Circle Transformation is both confusing and counterproductive
- I only looked at the pages linked from the lesson plans and it's a little tough to really weigh in before seeing the student-facing materials
- I sometimes had the sense that there was a lot of explaining rather than discovering, but I decided not to leave that feedback anywhere specific because my gut is that that feeling will likely go away once the student-facing materials exist.
- something about your use of the word equivalence feels imprecise to me, but I don't have a better suggestion or any concrete justification for why that is, so maybe that feeling isn't valid
- I'm not ready to weigh in on the sequencing of the lessons at this point until the student materials are there, but there isn't something glaringly out of order at this stage.

PR #1218, opened 2022-11-26
Revise Bar&Pie Chart lesson

### LINKED ISSUE #1217: Add stacked and multi bar charts
Should this be its own lesson, or fill the space left by "exploring displays" in the existing bar-and-pie-chart lesson? I'm leaning yes...

schanzer:
This PR closes #1217 , and includes the following changes:

- [ ] Bar and Pie Charts are now described in terms of "grouping by value"
- [ ] An explicit connection is made to the `count` function, which will be introduced in the new ds-contracts lesson
- [ ] Stacked and Multi Bar Charts are introduced as "grouping values within groups"

High-level question: _is this lesson too long?_ I can would imagine that some of the exercises here should be moved to "optional" if we want it to fit into a single lesson. Otherwise, this should be split into two lessons.


flannery-denny:
@schanzer let's talk about this. In my understanding I was going to introduce contracts for `count` and all one-variable displays in the functions for tables and displays lesson. To me that means that the focus of the bar and pie chart section should be on data science rather than pyret mechanics.
- why we would pick one over the other?
- what gets hidden in pie charts?
- emphasize the idea that bar charts work for any categorical data but pie charts only make sense if we know all of the categories. 
- the value of comparing a pie or bar chart about a population (for example school district demographics) to a control group (for example citywide demographics) - we have a file that focuses on this already for nyc. not sure why / how it got cut from the lesson.

If that's where you picked up with bar and pie charts, I think this could all fit into one lesson... and you could probably use the existing notice and wonders to hit some of the learning goals described above.

For some reason I don't have access to comment on the lesson plan itself through this pull request, but here are a few observations that could be addressed:
- The fourth lesson goal is a repeat - assuming you actually meant to rewrite it for multibar charts
- We should provide a starter file for pie-chart-summarized
- I don't understand what the data categories means for Refugees in the Americas
- blonde is spelled with an e

PR #1237, opened 2023-01-02
first draft of revised Table Functions lesson

schanzer:
This uses the new Shapes Table, plus a hidden `draw-shape` function to play around with.

Question: Is it worth bringing the Shapes Table back for the Composing-Table-Functions lesson? I could see pros and cons to that...

### DIFF
--- a/lessons/table-functions/langs/en-us/index.adoc
+++ b/lessons/table-functions/langs/en-us/index.adoc
@@ -63,19 +63,19 @@ Sometimes we want to *filter the rows* of a table by some criteria. In our anima
 
 [.lesson-roleplay]
 --
-- Select a volunteer who will play the role of… *Filter*!
+- Select a volunteer who will play the role of… `filter`!
 - Hand the volunteer a function card from the @handout{pages/function-cards.adoc, Function Cards} set. (Choose your students and cards carefully! Some cards only work if you have at least one student with glasses, red hair, etc.)
-- Instruct Filter to read the card carefully, making sure they understand what it will do for each "Row" (student!) in the "Table", but to __keep it a secret__.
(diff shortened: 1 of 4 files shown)


flannery-denny:
- [ ] adjust lesson goals away from Animals Dataset
- [ ] I believe this version of including a table is new to students at this stage. 
Screen Shot 2023-01-03 at 10 47 04 AM
Do you want to acknowledge that more explicitly in the lesson plan?
- [ ] Will these pyret messages already be familiar to students at this point in the curriculum or do we need to provide teachers with scaffolding about them?
Screen Shot 2023-01-03 at 10 46 56 AM
- [ ] If feels like the intent of these reading row and table function definition questions was to have some part of them be open-ended, but they aren't landing that way.
Screen Shot 2023-01-03 at 11 03 14 AM
- [ ] Reading Row and Table Function Definitions does not engage students with the "lookup-name" function found in the starter file.
- [ ] There is reference to the hidden function `draw-shape`. Do we want to let them know that it takes in a row?
- [ ] Have students seen an expression like this before? `draw-shape(row-n(shapes, 4))` I'm specifically curious about the `(shapes, 4)` part of it. Doesn't feel familiar to me.
- [ ] In the second section of the lesson plan, I think we need to come up with a way to consistently format all of the references to the Filter character.
- [ ] The table in the starter file is named "shapes". In the the Exploring Table Functions workbook page it is referred to as "shapes-table", which yields an error. Choose your preference and make sure that all code provided throughout the lesson plan is inconsistent.
- [ ] weird nit - when viewing the workbook page online I am able to copy the code and paste it into pyret... but when it gets to pyret... it was red dots that I need to delete to make the code usable. Not sure whether or not that's something worth addressing, but figured I'd note it. 
Screen Shot 2023-01-03 at 11 26 58 AM
- [ ] the workbook page directs students to the contracts page to find `build-column`... the contracts page lists `build-col` instead.

PR #1239, opened 2023-01-05
Early math 

retabak:
Interested in any and all thoughts and critiques that you have to offer, @flannery-denny !

Some notes on known issues:
- I haven't added definitions to the glossary yet, so you'll get warnings when you build.
- There are no slides yet (more warnings).
- There's an issue with worksheets that include Circles looking weird in the browser version (they wrap) but nice in the printed version. @schanzer is aware of it and working on it. 
- Language Table shows up (needlessly).

I remain excited about this pathway and my progress but suspect that there is heaps of work yet to be done!


lessons/Early-Math/translating/langs/en-us/pages/three-sets-cards.adoc
```diff
@@ -0,0 +1,25 @@
...
+</style>
+++++
+
+To prepare for this activity, give each pair of students an envelope containing three sets of cards. Keep each set of cards (Expressions, Circles of Evaluation, and Words) together with rubber bands or paperclips.
+
+[.FillVerticalSpace, cols="^.^1a,^.^1a,^.^1a", stripes="none", options="header"]
+|===
+|  Expressions		| Circles of Evaluation			| Words
+| @math{15+3}		| @show{(coe '(+ 15 3))}		| the sum of 15 and 3
+| @math{15 / 3}		| @show{(coe '(/ 15 3))}		| the quotient of 15 and 3
```
flannery-denny, line 15: While 15 / 3 looks more like what we see on the circle, in my experience, the students who will be engaging with this material will be far more comfortable with @math{15 \div 3} and, given the simplicity of the expressions being matched to circles it will be an opportunity for them to discover the connection between the symbols.

lessons/Early-Math/translating/langs/en-us/index.adoc
```diff
@@ -0,0 +1,183 @@
...
+
+=== Launch
+
+Give each pair of students an envelope and explain that it contains @handout{pages/three-sets-cards.adoc, "three sets of cards"}.
+
+[.lesson-instruction]
+- Look through the first set of cards. What do you notice?
+** _Students should notice that the first set of cards includes @vocab{mathematical expressions}. The expressions on these cards each include the number 15, the number 3, and an operator. Note that because these are expressions - not @vocab{equations} - they do not include an equal sign._
+- Now, look through the second set of cards. What do you notice?
+** _Students should observe that each card includes a circle. They may also notice the position of the numbers and operator within the circle._
```
flannery-denny, line 52: I think we can do better here. First, the circles of evaluation aren't actually circular, so I highly doubt that students will actually describe them as circles (unless you want to make the card sort a different way). Could we speak more broadly? I'm not sure I have the best words to offer up here, but perhaps the numbers are inside a diagram? or a shape?

lessons/Early-Math/translating/langs/en-us/index.adoc
```diff
@@ -0,0 +1,183 @@
...
+
+This second set of cards includes @vocab{_Circles of Evaluation_}. Circles of Evaluation help students visualize the structure of the mathematical expressions they encounter. For the time being, here’s what students need to know about them:
+
+- Every Circle must have one - and only one! - function, written at the top.
```
flannery-denny, line 56: Are you sure you want to talk about functions here? You could use operation instead, which is a word more commonly used in math books/conversations for this grade level. While they are functions in WeScheme, In pyret all of the things you are referring to as functions are categorized as operators...

lessons/Early-Math/translating/langs/en-us/index.adoc
```diff
@@ -0,0 +1,183 @@
...
+- One at a time, take turns reading each Words card out loud. After reading the card, place it with the appropriate Circle and Expression. Some sets will include just 3 cards, and others will have more.
+- Explain to your partner how and why you placed each card.
+- You and your partner must agree on each card’s placement before advancing to the next.
+
+Again, circulate during the activity. As students match expressions in words to their arithmetic expressions and Circles, additional challenges may arise. In particular:
+
+- *One-third of 15*. This card will likely motivate some discussion. Take advantage of opportunities to help students connect multiplication and division as inverse operations. Explain to students that another way to represent this expression is @math{\frac{1}{3} \times 15}. Dividing by 3 produces the same outcome as multiplying by @math{\frac{1}{3}}.
+- *Translating subtraction.* Students often translate "3 less than 15" into @math{3 - 15}, rather than @math{15 - 3}. Discuss the meaning of the word "less" in this context. Sometimes, asking, “What value is 3 less than 15?” can help students to make the connection.
```
flannery-denny, line 85: If students have already been introduced to inequality statements (3 < 5), there is often another point of confusion worth referencing here:
- "is less than" is a comparison statement (3 < 5)
- "less than" without the "is" indicates subtraction 
- "than" is a word that tends to force the numbers to appear in the opposite order in which the language writes about them
- "is" is a statement of inequality / translates to equals

lessons/Early-Math/translating/langs/en-us/pages/practice-translating.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+
+Each row represents a single arithmetic expression, written in three different forms. Fill in the empty spaces so that all three forms represent the same expression.
+
+[.FillVerticalSpace, cols="^.^1a,^.^4a,^.^4a,^.^4a", stripes="none", options="header"]
+|===
+| 	 | Expression in Words				| Circle of Evaluation		| Math Expression
+| @n | 24 increased by 3		 		| @show{(coe '(+ 24 3))}		| @ifsoln{@math{24 + 3}}
+| @n | @math{\frac {1}{3}} less than 4	| @showsoln{(coe '(- 4 1/3))}	| @math{4 - \frac {1}{3}}
+| @n | @ifsoln{the quotient of 24 and 8}| @showsoln{(coe '(/ 24 8))}	| @math{\frac {24}{8}}
```
flannery-denny, line 17: Probably better to use 24 \div 8 than \frac {24}{8}

lessons/Early-Math/translating/langs/en-us/pages/practice-translating.adoc
```diff
@@ -0,0 +1,25 @@
...
+| @n | the product of 9 and 11			| @showsoln{(coe '(* 9 11))}	| @ifsoln{@math{9 * 11}}
+| @n | @ifsoln{16 tripled}				| @show{(coe '(* 16 3))}		| @ifsoln{@math{16 \times 3}}
+| @n | half of 100						| @showsoln{(coe '(* 100 1/2))}	| @ifsoln{@math{\frac {1}{2} \times 100}}
+| @n | the difference between 20 and 8	| @showsoln{(coe '(- 20 8))}	| @ifsoln{@math{20 - 8}}
+| @n | @ifsoln{one-third of 4}			| @showsoln{(coe '(* 1/3 4))}	| @math{\frac{1}{3} \times 4}
+
+|===
+
```
flannery-denny, line 25: Noting that the trickiest translation concept you're introducing is statements using the word "than" and the only examples you've included with these already give the order of the numbers away. Suggest adding at least one more example where students are in charge of getting the order right based on the words.

lessons/Early-Math/translating/langs/en-us/index.adoc
```diff
@@ -0,0 +1,183 @@
...
+- On @printable-exercise{ambiguous-or-clear.adoc}, identify the expressions that have two different numeric translations.
+- When you encounter an expression that is ambiguous, rewrite it two times - once for each possible interpretation.
+- When you encounter an expression that is clear, draw its Circle of Evaluation.
+
+=== Synthesize
+
+- Why are some expressions in words ambiguous and others are not?
+- Do you think that expressions written in the language of math have ambiguity?
+- Are Circles of Evaluation ever ambiguous?
+
```
flannery-denny, line 183: Noting that, while @schanzer is fairly attached to this glossary definition,

> **circle of evaluation** a 'sentence diagram' of the structure of a mathematical expression

I, personally, think it's doing us a disservice because I love circles of evaluation and do not find sentence diagrams to be helpful or inspiring... nor do I think most students in the 21st century have any exposure to them.

Don't necessarily have energy to think this one through, and it's not a hill I have energy to defend, but encourage y'all to discuss. Any reason it can't just be described as a diagram (which most people recognize as a helpful tool)... rather than a sentence diagram?

lessons/Early-Math/translating/langs/en-us/pages/translation-table1.adoc
```diff
@@ -0,0 +1,22 @@
...
+++++
+
+
+Fill in any missing spaces on the table below.
+
+[.FillVerticalSpace, cols="^.^1a,^.^4a,^.^4a,^.^4a", stripes="none", options="header"]
+|===
+| 	 | Math 								| Circle of Evaluation				| Words
+| @n | @ifsoln{@math{5 + 3}}				| @showsoln{(coe '(+ 5 3))}			| Find the sum of 5 and 3
+| @n | @math{(5 + 3) \times 2}				| @showsoln{(coe '(* (+ 5 3) 2))}	| Find the sum of 5 and 3, then double it.
```
flannery-denny, line 16: Are you sure this is more aligned to materials students will see than "twice the sum of 5 and 3" or "double the sum of 5 and 3"?

Perhaps that materials I've seen are outdated, but I am unaccustomed to test and textbook questions splitting language that is to be translated into multiple sentences, as seems to be the norm on this entire page. (Parsing the order from your wording is a much simpler task... perhaps that's where math materials are headed?)

lessons/Early-Math/translating/langs/en-us/pages/translation-table2.adoc
```diff
@@ -0,0 +1,22 @@
...
+|===
+| 	 | Math 							| Circle of Evaluation							| Words
+| @n | @math{24 -6}						| @showsoln{(coe '(- 24 6))}						| @ifsoln{Find the difference between 24 and 6.}
+| @n | @math{24 - (6 \div 3)}			| @showsoln{(coe '(- 24 (/ 6 3)))}					| @ifsoln{Subtract the quotient of 6 and 3 from 24.}
+| @n | @math{(24 - 6) \div 3}			| @showsoln{(coe '(/ (- 24 6) 3))}					| @ifsoln{Find one-third of the difference of 24 and 6.}
+| @n | @math{\frac {1}{2} \times 100}	| @showsoln{(coe '(* 1/2 100))} 			| @ifsoln{Find half of 100.}
+| @n | @math{(\frac {1}{2} \times 100) + 10} |@showsoln{(coe '(+ (* 1/2 100) 10))}	| @ifsoln{Find the sum of 10 and half of 100.}
+| @n | @math{\frac {1}{2} \times (100 + 10)}	|@showsoln{(coe '(* 1/2 (+ 100 10)))} | @ifsoln{Find half the sum of 100 and 10.}
+|===
+
```
flannery-denny, line 22: I'm curious as to why you decided that page 1 would vary which representation was provided but page 2 would always provide words? (I don't necessarily object, just checking in to make sure it was intentional.)

lessons/Early-Math/translating/langs/en-us/pages/matching-math-to-words.adoc
```diff
@@ -0,0 +1,46 @@
...
+| 2 increased by the difference of 10 and one-third
+
+| @math{(2 + 10) - \frac {1}{3}}
+|*@n*@showsoln{"-D" }||*F*
+| the sum of 10 and one-third, doubled
+
+
+
+|===
+
```
flannery-denny, line 46: Most of the mathematical expressions included on this page use unnecessary parentheses that make it so that students do not have to think about whether addition or multiplication comes first. Is that intentional? It's something I've worked hard to avoid providing in materials I've developed.

lessons/Early-Math/computation/langs/en-us/index.adoc
```diff
@@ -0,0 +1,236 @@
...
+@keywords{computation, equivalence}
+
+[@lesson-intro-table]
+|===
+
+| Lesson Goals
+| Students will be able to...
+
+* Use computation to develop equivalent representations of a given expression.
+* Diagram solving with Circles using Computation.
```
flannery-denny, line 16: I am not able to parse what these goals are intended to mean.

flannery-denny: This lesson plan often uses "Circles" in place of "Circles of Evaluation". I highly encourage you to use all of the words. I made the change for you in the first lesson and there were only one or two instances in the entire lesson plan where using just the word circle made sense to me and it was because you were referring to the shape itself rather than the shape and the numbers and the operator.

lessons/Early-Math/computation/langs/en-us/pages/circles-and-computation.adoc
```diff
@@ -0,0 +1,51 @@
...
+|    | Circle Solving Diagram
+| @n | @show{(coe coe1a)} &rarr; @show{(coe coe1b)} &rarr; @math{30}
+| @n | @show{(coe coe2a)} &rarr; @show{(coe coe2b)} &rarr; @math{36}
+| @n | @show{(coe coe3a)} &rarr; @show{(coe coe3b)} &rarr; @math{4}
+| @n | @show{(coe coe4a)} &rarr; @show{(coe coe4b)} &rarr; @show{(coe coe4c)} &rarr; @math{25}
+| @n | @show{(coe coe5a)} &rarr; @show{(coe coe5b)} &rarr; @show{(coe coe5c)} &rarr; @math{27}
+| @n | @show{(coe coe6a)} &rarr; @show{(coe coe6b)} &rarr; @show{(coe coe6c)} &rarr; @math{12}
+|===
+
+
```
flannery-denny, line 51: In some math curricula (cmp for example) circles of evaluation might be useful well before students learn to multiply fractions or decimals. And since working with fractions and decimals are also often introduced in two separate units, you might even want to disaggregate them.

I would suggest making multiple versions of this workbook page: 1 with whole numbers, others with fractions/decimals/integers.

lessons/Early-Math/computation/langs/en-us/pages/true-or-false.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+++++
+<style>
+div.circleevalsexp { width: auto; }
+td > .content > .paragraph > * { vertical-align: middle; }
+</style>
+++++
+
+
+Fill in any missing spaces on the table below.
```
flannery-denny, line 11: These directions could be much stronger.

lessons/Early-Math/computation/langs/en-us/index.adoc
flannery-denny: In general, please use Circles of Evaluation instead of Circles almost all of the time. I have now updated two entire lesson plans and have found only a handful of spots where the word circle made sense in lieu of Circles of Evaluation.

lessons/Early-Math/commutative-property/langs/en-us/pages/discover-comm.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+++++
+<style>
+  table {grid-template-rows: 3fr 1fr !important;}
+  div.circleevalsexp .value,
+  div.circleevalsexp .studentBlockAnswerFilled { min-width:unset; }
+</style>
+++++
+
+@n Analyze and evaluate the two Circles below:
```
flannery-denny, line 11: I understand the evaluate direction, but I am not sure what you're asking me to do with the word Analyze.

lessons/Early-Math/commutative-property/langs/en-us/pages/discover-comm.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+[.FillVerticalSpace, cols="^.^3,^.^3"]
+|===
+|@showsoln{(coe '(* 4 3))}	| @showsoln{(coe  '(* 3 4))}
+| @ifsoln{@math{4 \times 3 = ?}} | @ifsoln{@math{3 \times 4 = ?}}
+|===
+
+
+
+@n Analyze and evaluate the Circles below to decide if the Commutative Property holds for problems involving *division*.
```
flannery-denny, line 41: ibid re: analyze

lessons/Early-Math/commutative-property/langs/en-us/pages/discover-comm-cont.adoc
```diff
@@ -0,0 +1,68 @@
...
+</style>
+++++
+
+@n Analyze and evaluate the Circles below to decide if the Commutative Property holds for problems involving *addition*.
+
+@vspace{1ex}
+
+[.FillVerticalSpace, cols="^.^3,^.^3"]
+|===
+|@show{(coe '(+ (+ 9 2) 8))}	| @show{(coe  '(+ (+ 8 2) 9))}
```
flannery-denny, line 17: I would describe the example here as illustrating the Associative property because you have grouped things differently.

Strongly suggest using (9+2)+8 and 8 + (9+2) instead.

lessons/Early-Math/commutative-property/langs/en-us/pages/discover-comm-cont.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+++++
+<style>
+  table {grid-template-rows: 3fr 1fr !important;}
+  div.circleevalsexp .value,
+  div.circleevalsexp .studentBlockAnswerFilled { min-width:unset; }
+</style>
+++++
+
+@n Analyze and evaluate the Circles below to decide if the Commutative Property holds for problems involving *addition*.
```
flannery-denny, line 11: I still don't know how to analyze a circle.

lessons/Early-Math/commutative-property/langs/en-us/pages/comm-prop-table.adoc
```diff
@@ -0,0 +1,55 @@
...
+div.circleevalsexp { width: auto; }
+</style>
+++++
+
+
+For each Circle, apply the Commutative Property as many times as you can.
+
+[.FillVerticalSpace,cols=".^1a,^.^6a,^.^6a,^.^6a,^.^6a", stripes="none", options="header"]
+|===
+|	 | Analyze this Circle. | Apply the Commutative Property. | Apply it again. | Apply it again?
```
flannery-denny, line 15: What do you actually want them to do in the part that says Analyze this Circle. 
I would probably label this column "Begin with this Circle of Evaluation" or "Original Expression"

lessons/Early-Math/commutative-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+Explain to your students that Claire and Walker are both translating expressions in words into Circles of Evaluation. Their teacher asks them to represent the following expression: "one half the product of five and six."
+
+[.embedded, cols="^.^1,^.^1", grid="none", stripes="none" frame="none"]
+|===
+
+| Claire draws this:				| Walker draws this:
+|@show{(coe  '(* 1/2 (* 5 6)))}		| @show{(coe  '(* (* 5 6) 1/2))}
+|===
+
```
flannery-denny, line 93: I'm not sure this is the most interesting discussion.

As a teacher interested in getting my students to fluency and precision with writing code, I saw the commutative property lead to many errors as most students who didn't stick to the order didn't do so because they were thinking about whether it was addition/multiplication or subtraction/division. They just either followed the order or they didn't. 

A much more important and interesting discussion I would want my students to debate is about the student who write the circle (/ (*5 6) 2), which I would see as an equally excellent (if not better) way of translating those words.

lessons/Early-Math/commutative-property/langs/en-us/pages/claire-and-walker.adoc
```diff
@@ -0,0 +1,50 @@
...
+| @show{(coe '(- 6 (* 3 20)))}
+| @ifsoln{Claire's Circle is correct. Subtraction is not commutative.}
+
+
+| @n
+| The product of 4 and the difference of 3 and 1.
+| @show{(coe '(* 4 (- 1 3)))}
+| @show{(coe '(* 4 (- 3 1)))}
+| @ifsoln{Walker's Circle is correct. Subtraction is not commutative; "the difference of 3 and 1" must be translated as @math{3 - 1}.}
+|===
```
flannery-denny, line 50: How about including an example that reinforces that division is not commutative?

lessons/Early-Math/commutative-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,120 @@
...
+- In the column on the right, record if Claire, Walker, or both students correctly translated the words into a Circle.
+- Optional: Complete @printable-exercise{words-to-circles-challenge.adoc} to analyze Circles of Evaluation and commutativity for a more complex expression in words.
+
+Discuss and debrief with students. Invite students to verbally share their responses to reinforce important vocabulary and concepts that students will use again and again in future lessons.
+
+=== Synthesize
+
+Do you translate words into Circles precisely and directly – or do you sometimes make changes to your Circles as you translate? Why?
+
+If you were a teacher, would you require _exact_ translations of circles, or would you allow students to modify their Circles when translating?
```
flannery-denny: I am curious why this is the most important point you want to make in a lesson that is introducing the commutative property.  Some teachers will feel very strongly about landing on one side or another of this conversation. I think there are probably many other ideas that all teachers and students could agree are important takeaways from this lesson...

lessons/Early-Math/associative-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+* Let's explore the Associative Property via Circles of Evaluation.
+
+| Materials
+|[.materials-links]
+@material-links
+
+| Key Points For The Facilitator
+|
+* Just as when we introduced the Commutative Property, we asked students exclusively to manipulate the order of the values... when we introduce the Associative Property, we will ask students exclusively to manipulative the groupings of the values.
```
flannery-denny, line 29: This is a mouthful and punctuation is unconventioinal. Suggest rewriting.

lessons/Early-Math/associative-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+== The Associative Property @duration{30 minutes}
+
+=== Overview
+
+Students analyze pairs of Circles to see if they represent equivalent expressions. The Circles make visible the structural changes that occur when we apply the Associative Property.
+
+=== Launch
+
+The Associative Property defines an equivalence relation over structures, such that "adding 4 to the sum of 3 and 2" is equivalent to "adding the sum of 4 and 3 to 2." In the short term, familiarity with the Associative Property enables students to make mental computations more efficiently. In the long term, comfort with the laws of arithmetic will provide a strong foundation for more complex algebraic reasoning.
```
flannery-denny, line 41: I think most people will have to read this more than once in order to make sense of what you're trying to say. Any good reason not to use the words "grouping" or "regrouping" before the concrete example to give teachers a heads up of what they're looking for in the examples?

retabak: Ack! I could've sworn I changed this! It is definitely not written in a student- or teacher-friendly way.

flannery-denny (commented):
- [ ] As discussed in previous pull request, "Circles of Evaluation" should really be referred to using their full name. "Circles" on its own doesn't actually mean anything. I have made copious changes. I am sure I've missed many and you will write new directions after this request. Please finish the cleanup.
- [ ] Figure out what directions you want to use on workbook pages instead of "Analyze" and deploy throughout.
- [ ] Deploy `(?ANS)` throughout workbook pages.
- [ ] FYI - A number of workbook pages are currently 2 pages long

> WARNING: in early-math workbook-long.pdf, ../../lessons/identity-property/pages/discover-identity.pdf has 2 pages
> WARNING: in early-math workbook-long.pdf, ../../lessons/identity-property/pages/which-one-doesnt-belong.pdf has 2 pages
> WARNING: in early-math workbook-long.pdf, ../../lessons/absolute-value-and-opposites-2/pages/three-sets-cards.pdf has 2 pages
> WARNING: in early-math workbook-long-sols.pdf, ../../lessons/associative-property/solution-pages/discover-assoc.pdf has 2 pages
> WARNING: in early-math workbook-long-sols.pdf, ../../lessons/associative-property/solution-pages/discover-assoc-cont.pdf has 2 pages
> WARNING: in early-math workbook-long-sols.pdf, ../../lessons/identity-property/solution-pages/discover-identity.pdf has 2 pages
> WARNING: in early-math workbook-long-sols.pdf, ../../lessons/identity-property/solution-pages/which-one-doesnt-belong.pdf has 2 pages
> WARNING: in early-math workbook-long-sols.pdf, ../../lessons/absolute-value-and-opposites-2/solution-pages/three-sets-cards.pdf has 2 pages

flannery-denny (commented):
As discussed in our call, I'd suggest a rewrite of the Associative Property Lesson:
- [ ] The Associative Property lesson would be much more powerful if you focused on the power of regrouping rather than the mechanics of its possibility. 2 X 17 X 5 can be done in lots of orders... but if a student notices that they could group 2 x 5 = 10 it becomes a simple mental math problem... and, to me, this is the essence of why a student should care about the property!
- [ ] My understanding of the associative property is fairly aligned with this definition
> Associative property states that the sum or product of three or more numbers remains the same, irrespective of how they are grouped
- [ ] The current distinction between which problems can be solved with the Associative property  and which require both associative and commutative feels narrow... and leaves kids (and teachers!) focused on definitions rather than giving them permission to think flexibly.
- [ ] I'm not sure circles of evaluation do as good of a job of making this regrouping capacity visible as other connector symbols... which makes me think we may have to also show other representations and address that concern (are we regrouping or reordering?) explicitly. 
- [ ] my most concrete suggestion is to start the problems with expressions rather than circles of evaluation. Have students draw all possible coe. And then make sure to ask them to reflect on which makes their work most efficient.
- [ ] might be worth discussing in our next curriculum meeting?

Absolute Value & Opposites:
- [ ] I am not totally convinced by these lessons... the card sort (perhaps there should also be a Desmos version of this?) and matching circles and expressions page (presented kind of as an after thought) feel most compelling...
- [ ] I have concerns about the dominant role that the `negate` and `num-abs` functions are playing in these lessons given that they don't exist outside of our materials. Want to make sure we a) are providing enough support in how we introduce the named functions in this pathway b) are making sure to transfer understanding they are generating back to universally communicable math understanding.

Overarchingly:

- [ ] As discussed in previous pull request, "Circles" doesn't have any concrete meaning related to this work. I have pushed a lot of changes to reword to "Circles of Evaluation". I expect I missed some and that you will keep writing things. Please orient toward using the full phrase.
- [ ] In general students this age are unlikely to have experience with the word functions and it probably makes more use to talk about operations or operators (which aligns with Pyret).
- [ ] You have a lot of student-facing directions on workbook pages asking them to analyze things. I am not sure what I am supposed to do to satisfy that direction. My guess is that, while analyze might be something a teacher suggests students do, a workbook page should never use analyze in the directions, opting for more concrete deliverables instead such as evaluate.
- [ ] I have made some changes to get rid of the word "worksheet". We don't refer to any of our materials as worksheets. Use "workbook page" whenever necessary to generalize.
- [ ] Figure out what directions you want to use on workbook pages instead of "Analyze" and deploy throughout.
- [ ] Deploy `(?ANS)` throughout workbook pages.
- [ ] Scale back any worksheet that is currently more advanced to focus on whole numbers and make an issue adding intentions for plans to add more advanced material in the future
- [ ] Some of these lesson never refer to the word expression and only refer to Circles of Evaluation. I think it would behoove us to make sure that we are thinking COE as a tool rather than presenting it as the sole representation of mathematical expressions.
- [ ] I am a little uncertain about our use of the word equivalence. I think of 5(2+5) as being equivalent to 5x7 and 10 + 25 and 35, but not to 30 + 5, although it evaluates to the same thing. I am not sure where the canon lands on that.
- [ ] Noting that this pathway does not make much use of Notice & Wonder pedagogy and could.
- [ ] FYI - A number of workbook pages are currently 2 pages long - **RT needs @schanzer's help to fix the one that remains.** 

> WARNING: in early-math workbook-long-sols.pdf, ../../lessons/identity-property/solution-pages/discover-identity.pdf has 2 pages

lessons/Early-Math/associative-property/langs/en-us/pages/assoc-prop-table.adoc
```diff
@@ -0,0 +1,52 @@
...
+div.circleevalsexp { width: auto; }
+</style>
+++++
+
+
+For each Circle, apply the Associative Property as many times as you can. Some Circles have several equivalent expressions, while some have none.
+
+[.FillVerticalSpace,cols=".^1a,^.^6a,^.^6a,^.^6a,^.^6a", stripes="none", options="header"]
+|===
+|	 | Analyze this Circle. | Apply the Associative Property. | Apply it again. | Apply it again?
```
flannery-denny, line 14: analyze

flannery-denny: How are you expecting us to apply the associative property? As another circle? As an expression with parentheses added?

flannery-denny: I honestly wonder whether this page would work better if you started with an expression and asked students to draw as many circles of evaluation as they could by applying the associative property.... and perhaps you could also ask which would make it easiest to evaluate the expression (you might want to choose your numbers differently to support that learning goal)

lessons/Early-Math/associative-property/langs/en-us/pages/true-or-false.adoc
```diff
@@ -0,0 +1,22 @@
...
+</style>
+++++
+
+Is the equation represented by the two Circles true or false? Explain why, and then describe what you would change to make the _false_ equations _true_.
+
+[.FillVerticalSpace,cols="^.^1a,^.^8a,^.^5a", stripes="none", options="header"]
+|===
+|	 | Circles.																	   |
+True or False? Explain.
+| @n |@show{(coe '(* (* 8 9) 10))}	@math{=} @show{(coe '(* (* 8 9) 10))}  |
```
flannery-denny, line 16: confirming you meant for the first 2 to be exactly the same. I see pros and cons to this - not objecting. just making sure it was intentional

lessons/Early-Math/associative-property/langs/en-us/pages/true-or-false.adoc
```diff
@@ -0,0 +1,22 @@
...
+td > .content > .paragraph > * { vertical-align: middle; }
+</style>
+++++
+
+Is the equation represented by the two Circles true or false? Explain why, and then describe what you would change to make the _false_ equations _true_.
+
+[.FillVerticalSpace,cols="^.^1a,^.^8a,^.^5a", stripes="none", options="header"]
+|===
+|	 | Circles.																	   |
+True or False? Explain.
```
flannery-denny, line 15: I wonder if you could make the table header incorporate the full set of directions?
True? Explain.
False? What you need to change to make it True?

lessons/Early-Math/associative-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+== Circles, the Associative Property, and Mental Math @duration{20 minutes}
+
+=== Overview
+Students discover how the Associative Property and Circles make mental computation simpler.
+
+=== Launch
+
+[.lesson-instruction]
+- Alecia and Jill are trying to solve this multiplication problem: @math{4 \times 36 \times 2.5}
```
flannery-denny, line 74: Flagging that your example uses decimals.

lessons/Early-Math/equivalence/langs/en-us/index.adoc
```diff
@@ -0,0 +1,166 @@
...
+** _Possible responses: The same numbers and operations are included. This time, the product (or the inner circle) and @math{4} swap positions._
+- Which is true: Equation 1, Equation 2, or both? How do you know?
+** _The first equation is true. We see the Commutative Property of multiplication being applied when @math{22 \times 3} is transformed to @math{3 \times 22}._
+** _In the second equation, it looks as though there has been a misguided attempted to apply the Commutative Property; this does not work because the Commutative Property does not hold for subtraction._
+** _We can also use Computation to verify that both sides of the first equation equal @math{62}. The second equation's left side is @math{62}, but its right side is @math{-62}._
+
+As students observe and share out, urge them to justify their ideas about equivalence.
+
+=== Investigate
+
```
flannery-denny, line 72: Suggest overhauling these instructions so that they are something more in the direction of:

We have learned that computation is an effective strategy for determining whether or not two expressions are equivalent. Pretty soon we'll be looking at expressions with variables, where our computation strategy won't work as well. Let's see if we can figure out how to determine whether or not two expressions are equivalent without computation! For now, while we're still looking at expressions with numbers, we can doublecheck our thinking with computation.

lessons/Early-Math/translating/langs/en-us/index.adoc
retabak: Hm. This is interesting. I'm not sure I want to talk about functions - but I wonder if we want to preserve that language for when students start using abs-val and negate? 

I'm currently leaning toward putting: "Every Circle must have one - and only one! - function (in this case, an _operator_), written at the top."

What do you think? Maybe @schanzer wants to weigh in?

... I should probably use @vocab to add "function" and "operator" to the Early Math glossary.

lessons/Early-Math/translating/langs/en-us/index.adoc
retabak: this is great! Thanks.

lessons/Early-Math/equivalence/langs/en-us/pages/which-one-doesnt-belong.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+| @n
+| @show{(coe '(- ( + (* 8 10) 6) 5))}
+@show{(coe '(- (+ 6 (* 8 10)) 5))}
+@show{(coe '(- (+ (* 10 8) 6) 5))}
+@show{(coe '(- 5 (+ (* 8 10) 6)))}
+| @ifsoln{The fourth Circle does not belong. It shows @math{5 - 86} rather than @math{86 -5}. The Commutative Property does not hold for subtraction.}
+
+|===
+
```
flannery-denny, line 61: love this as a concept

lessons/Early-Math/identity-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+== The Identity Property
+
+=== Overview
+
+Students analyze pairs of Circles to see if they represent equivalent expressions. The Circles make visible the structural changes that occur when we apply the Identity Property.
+
+=== Launch
+
+Build on prior knowledge by reminding students of the activity that they completed in the previous lesson. You might even refer to the list of category names that the class created, and then draw students’ attention to any category names that allude to or suggest the commutativity. (Any mention of “changing the order” is a strong hint that students’ have clued into the Commutative Property!)
```
flannery-denny: guessing you meant to give a clue for the Identity property rather than the commutative property here?

lessons/Early-Math/identity-property/langs/en-us/pages/discover-identity.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+@fitb{}
+
+
+@vspace{1ex}
+
+These Circles represent the Identity Property!
+
+- The Identity Property of Multiplication tells us that the product of 1 and any number is that number.
+
```
flannery-denny, line 36: Could also be articulated as "tells us that multiplying 1 does not change the value."

lessons/Early-Math/identity-property/langs/en-us/pages/discover-identity.adoc
```diff
@@ -0,0 +1,86 @@
...
+- The Identity Property of Addition tells us that the sum of a number and zero is that number.
+
```
flannery-denny, line 38: Could also be articulated as "tells us that adding 0 does not change the value."

lessons/Early-Math/identity-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+[.lesson-instruction]
+- Now, turn to @printable-exercise{discover-identity.adoc}. Look at the first problem. Translate the expressions in words into Circles of Evaluation. What do you Notice? What do you Wonder?
+** _Possible responses: All three Circles include @math{12 + 4}. All three Circles evaluate to @math{16}._
+- Complete the worksheet by filling in the blanks for the remaining Circles. Avoid putting the same operator more than once.
+- Keisha says that when she multiplies a number by zero, she always gets that number. Do you agree?
```
flannery-denny, line 46: I'm not seeing what this reference to Keisha's thinking is referring to...

lessons/Early-Math/identity-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,108 @@
...
+** _No, multiplying by zero produces zero._
+- What did you observe about the Identity Property?
+** _Possible responses: I can multiply or divide by @math{1} and get an equivalent result. I can add or subtract @math{0} and get an equivalent result._
```
flannery-denny, line 49: You're mentioning division and subtraction here. Feels like that comes out of nowhere...

lessons/Early-Math/identity-property/langs/en-us/pages/discover-identity.adoc
```diff
@@ -0,0 +1,86 @@
...
+@fitb{}
+
+@fitb{}
+
+
+@vspace{1ex}
+
+@n Fill in the blank spaces below to create four unique representations of the Identity Property. Use each operation (addition, subtraction, multiplication, and division) at least once.
+
```
flannery-denny, line 62: Perhaps add a prethink here? "How might the identity properties for addition and multiplication connect to division and subtraction? Test your thinking out be completing the Circles of Evaluation below."

lessons/Early-Math/translating/langs/en-us/pages/practice-translating.adoc
retabak: I actually think so important for teachers vary their representation of division, largely to help students feel more comfortable with fractions. Perhaps I should put a note indicating this? Not recognizing fractions as division can be a big barrier to student success in middle school math.

retabak: Fifth grade CCSS standard - "Interpret a fraction as division of the numerator by the denominator (a/b = a ÷ b)."

lessons/Early-Math/translating/langs/en-us/pages/translation-table1.adoc
retabak: Okay, your question prompted me to do some research.  The Common Core standards clearly use phrases like the ones I've included.  I then went and looked at practice LEAP tests to see what sort of phrases were used there. In fifth grade, the relevant question uses both forms. In sixth grade, the relevant question uses the phrasing that you are more familiar with.

And so, I'm not sure. I suppose maybe I should include a mix? Or scaffold them, assuming that it's easier when the wording is given in direction form?

[Here are the results of my research, if you're interested.](https://docs.google.com/document/d/1yrdJxf1DbE_U5i6l5nJd1zfURuDNtE5TTVpOUdzCxBk/edit?usp=sharing)

lessons/Early-Math/translating/langs/en-us/pages/translation-table2.adoc
retabak: Jumping between representations exercises a different muscle than always going in the same direction (math, to Circles, to words - as on the 2nd worksheet); the first worksheet also ensures that students are exposed to example responses of all the different representations.

(I do suspect that I had a more concrete reason for designing the worksheets in this way - it was definitely intentional! - but I can't quite put my finger on it now.)

lessons/Early-Math/translating/langs/en-us/pages/matching-math-to-words.adoc
retabak: I will fix! Thanks.

lessons/Early-Math/identity-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,108 @@
...
+|@show{(coe '(+ 3 q))}	| @math{=} | @show{(coe '(* (+ 3 q) 1))}
+| @math{3 + q} 	| @math{=} | @math{(3 + q) \times 1}
+|===
+
+Or like this:
+
+[.embedded, cols="^.^3,^.^1,^.^3", grid="none", stripes="none" frame="none"]
+|===
+|@show{(coe '(+ 3 q))}	| @math{=} | @show{(coe '(+ 0 (+ 3 q)))}
+| @math{3 + q} 			| @math{=} | @math{(3 + q) + 0}
```
flannery-denny, line 87: you've changed the order from the expression to the circle here... suggest being consistent

lessons/Early-Math/identity-property/langs/en-us/pages/true-or-false-variables.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+
+
+| @n
+| @show{(coe '(- 76.42 (- (* 5 k) (* 5 k))))}
+@math{=}
+@show{(coe '(- 76.42 (- (/ w 3) (/ w 3))))}
+|
+
+
```
flannery-denny, line 61: While I recognize that you're showing equivalence here, to me only number 5 feels like it's representing the identity property

lessons/Early-Math/identity-property/langs/en-us/pages/which-one-doesnt-belong-variables.adoc
```diff
@@ -0,0 +1,80 @@
...
+@show{(coe '(+ (+ a m) (/ t t)))}
+|
+
+
+| @n
+| @show{(coe '(/ (/ h h) (+ m 2)))}
+@show{(coe '(/ (+ 2 m) (/ p p)))}
+@show{(coe '(/ (+ m 2) (/ g g)))}
+@show{(coe '(* (/ c c) (+ 2 m)))}
+|
```
flannery-denny, line 67: I think all of these belong?

lessons/Early-Math/translating/langs/en-us/pages/practice-translating.adoc
flannery-denny: I totally agree that it's important for students to recognize fractions as a division. By using `/` instead of `\div` you are losing an opportunity to connect the two!

lessons/Early-Math/distributive-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+* Let's explore the Distributive Property and Circles of Evaluation.
+
+| Materials
+|[.materials-links]
+@material-links
+
+| Key Points For The Facilitator
+|
+* A common approach to teaching the Distributive Property is by using an area model. This approach works well for students who are comfortable visualizing and computing area - but may leave other students confused.
+* We introduce the Distributive Property by inviting students to look at patterns in _structure_, and then later apply what they know using the area model.
```
flannery-denny, line 32: This second bullet describes what we're doing. 
I'd suggest leaving out the first bullet. If it feels critical to you, I'd suggest rewording so that it feels like what we're doing is additive, rather than suggesting that it is a mistake to use the area model.

lessons/Early-Math/distributive-property/langs/en-us/pages/sum-to-product.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+@vspace{1ex}
+
+[cols="^.1,^.^6,^.^2,^.^6", grid="none", frame="none", stripes="none"]
+|===
+| @n
+|@show{(coe '(+ (* 4 10) (* 4 9)))}
+| &rarr;
+| @show{(coe '(* 4 (+ 10 9)))}
+|| @fitb{10em}{@math{40 + 36}} | | @fitb{10em}{@math{4 \times 19}}
```
flannery-denny, line 20: Might make more sense to have these say 40 + 36 = 76 and 4 x 19 = 76?

lessons/Early-Math/translating/langs/en-us/pages/practice-translating.adoc
retabak: I'm not using `/` here! I'm using `\frac` !

lessons/Early-Math/distributive-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+** _Allow students to share their approaches. The Circles are equivalent - but Elise's explanation is not complete._
+- Try using computation to evaluate each Circle. What do you notice?
+** _Sample responses: The Circles are equivalent. They both evaluate to 414. To evaluate the first Circle, I found a *product*. To evaluate the second Circle, I found a *sum*._
+--
+
+Share with students that the Circles are equivalent because of the @vocab{Distributive Property}, whereby @math{a \times (b + c) = ab + ac}. Facilitate a discussion where students think deeply about what is happening in the scenario above, namely, that distribution allows us to rewrite a product as a sum or a sum as a product.
+
+[.lesson-point]
+Multiplying the sum of two addends by a number produces the same result as multiplying _each_ addend by that number and then adding the products.
+
```
flannery-denny, line 67: I wonder whether this explanation might be more digestible without the word addend?

> Multiplying the sum of two numbers by a third numbers produces the same result as multiplying each of the numbers by the third number before finding the sum.

lessons/Early-Math/distributive-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+=== Launch
+
+Elise and Dani are discussing the two Circles, below.
+
+[.embedded, cols="^.^3,^.^3", grid="none", stripes="none" frame="none"]
+|===
+| @show{(coe '(* 9 (+ 40 6)))} | @show{(coe '(+ (* 9 40) (* 9 6)))}
+|===
+
```
flannery-denny, line 50: I wonder if this launch would be better served with a notice and wonder than a debate? I usually use debates between two characters as something for students to respond to in writing on assessments and homework assignments rather than as openers.

lessons/Early-Math/commutative-property/langs/en-us/pages/discover-comm-cont.adoc
retabak: The bottom half of the worksheet is subtraction? (And always has been? Maybe it didn't build for some reason? Hoping you can see it.

(Agree with other feedback and responding to it.)

lessons/Early-Math/distributive-property/langs/en-us/pages/sum-to-product.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+| @n
+|@show{(coe '(- (* 8 206) (* 8 6)))}
+| &rarr;
+| @show{(coe '((?ANS *) (?ANS 8) ((?ANS -) (?ANS 206) (?ANS 6))))}
+|| @fitb{10em}{@ifsoln{@math{1648 - 48}}} | |  @fitb{10em}{@ifsoln{@math{8 \times 200}}}
+|===
+
+
+
```
flannery-denny, line 58: flagging use of decimals.

lessons/Early-Math/commutative-property/langs/en-us/pages/discover-comm-cont.adoc
retabak: Tried to implement this feedback. Take a look and see if it feels better to you!

lessons/Early-Math/distributive-property/langs/en-us/pages/mental-math.adoc
```diff
@@ -0,0 +1,54 @@
...
+| @math{210}
+
+| @n
+| @show{(coe '(* 20 29))}
+| @show{(coe '(* (?ANS 20) (- (?ANS 30) (?ANS 1))))}
+| @show{(coe '(- (* (?ANS 20) (?ANS 30)) (* (?ANS 20) (?ANS 1))))}
+| @ifsoln{@math{580}}
+
+| @n
+| @show{(coe '(* 50 49))}
```
flannery-denny, line 29: let's change this one to 50 X 51. so you've practiced both adding and subtracting with 1 before switching distance.

lessons/Early-Math/distributive-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+If we were asked to compute @math{70 \times 39}, we could do so using a calculator or the standard algorithm. Now that we are familiar with the Distributive Property, we have a valuable tool for efficient mental computation.
+
+[.lesson-instruction]
+- Turn to @printable-exercise{mental-math.adoc} and look at the first problem.
+- Our goal here is to make the math easier by creating an equivalent expression that we can solve in our heads! So, instead of finding the product of 70 and 39, we are going to multiply 70 by the difference of 40 and 1.
+- Complete the next Circle, which shows that we are going to _distribute_ 70.
+- We can now compute our solution by finding difference between two products.
+- Do you find this process more efficient than the standard algorithm for multiplication? Explain.
+** _Note: Without ample practice, there is a good chance that students will not find this process efficient. Like any solving strategy, it takes practice!_
```
flannery-denny, line 145: This might be a good opportunity to encourage teachers to have their students write 5 problems that this strategy would work well with. Alternatively those directions could be written as a challenge at the end of the workbook page using @star.

lessons/Early-Math/distributive-property/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+
+=== Launch
+
+[.lesson-instruction]
+--
+- Look at this expression: @math{45 \times 81 - 45 \times 79}
+- Can we simplify this to an equivalent expression? How?
+** _We can apply the Distributive Property, turning this into @math{45 \times (81 - 79)}._
+- Can you represent your solving process with Circles?
+** _Invite a student to draw a Circle representation on the board. See below for a possible Circle representation._
```
flannery-denny, line 121: Not sure how to reword this but we don't need to introduce the new term Circle representation on top of Circles of Evaluation.

lessons/Early-Math/absolute-value-and-opposites/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+* Let's use Circles of Evaluation to think about opposites and absolute value.
+
+
+| Materials
+|[.materials-links]
+@material-links
+
+| Preparation
+|
+* Draw or project a number line, which will be the focal point of a lot of the conversation for today's lesson.
```
flannery-denny, line 31: Let's be more specific! Of the integers from -10 to 10? or are you picturing something else?

lessons/Early-Math/absolute-value-and-opposites/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+| Student-facing Lesson Goals
+|
+
+* Let's use Circles of Evaluation to think about opposites and absolute value.
+
+
+| Materials
+|[.materials-links]
+@material-links
+
```
flannery-denny, line 28: Screen Shot 2023-01-16 at 1 48 21 PM
Do you prefer ampersand or _and_? Be consistent.

lessons/Early-Math/absolute-value-and-opposites/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+[.lesson-instruction]
+- What number is the same distance from zero as @math{8}?
+** _@math{-8}_
+- What number is the same distance from zero as @math{-3}?
+** _@math{3}_
+- The pairs of numbers we've just identified are called @vocab{opposites} - two numbers that are the same distance from zero on the number line, with one negative and one positive.
+- What are some other pairs of opposites that you can identify?
+** _Possible responses: @math{6.5} and @math{-6.5}; @math{4} and @math{-4}; @math{\frac {7}{3}} and @math{- \frac {7}{3}}._
+- Are there any numbers on this line that don't have an opposite?
+** _Students may guess that zero does not have an opposite - however zero is its own opposite!_
```
flannery-denny, line 58: I'm not sure that this is a point worth focusing on... it just feels like it's meant to trick students without actually deepening understanding. Why does it matter whether zero has no opposite or is its own opposite? The only answer I can come up with is if our understanding of opposites is that they are additive inverses... which is a concept you haven't introduced.

lessons/Early-Math/absolute-value-and-opposites/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+
+=== Investigate
+
+Explain to students that the function used to represent "the opposite" is called `negate`. It can be used like any other operator that we put at the top of a Circle of Evaluation - with one important difference: when we use `negate`, we can include just one argument inside the Circle.
+
```
flannery-denny, line 63: This is the first reference to a named function in the entire pathway! Are you expecting students to be hands on keys in this lesson? If so, we went to introduce the editor! If not, why are you introducing a named function?

lessons/Early-Math/absolute-value-and-opposites/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+== Absolute Value @duration{30 minutes}
+
+=== Overview
+
+Students consider the meaning of absolute value, and apply the concept to Circles of Evaluation using `num-abs`.
+
+=== Launch
+
+[.lesson-instruction]
+What is the distance between these two points on the number line: @math{-18} and @math{5}?
```
flannery-denny, line 116: We have asked them to draw a number line of the integers from -10 to 10. Either pick two integers within that range or change the directions for which numbers should be on the numberline that students and teachers draw.

lessons/Early-Math/absolute-value-and-opposites/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+| Key Points For The Facilitator
+|
+* The difference between absolute value and opposite can be subtle and confusing for students. Sometimes, given an input, they produce the same output - but at other times, their outputs differ.
+|===
+
+== Finding Opposites @duration{20 minutes}
+
+=== Overview
+
+Students discover what it means for numbers to be @vocab{opposites}, and explore the impact of the `negate` in Circles of Evaluation.
```
flannery-denny, line 42: Not sure what this sentence is supposed to mean.

lessons/Early-Math/absolute-value-and-opposites/langs/en-us/index.adoc
```diff
@@ -0,0 +1,150 @@
...
+
+Because _opposites_ are the same distance away from zero, they will always have the same absolute value. So, @math{|4| = 4} and @math{|-4| = 4}.
+
+
+=== Investigate
+
+The function to represent absolute value is `num-abs`. It can be used like any other operator that we put at the top of a Circle of Evaluation - with one important difference: when we use `negate`, we include just one argument inside the Circle.
+
+[.lesson-instruction]
+- Compare expressions with `num-abs` to expressions with `negate` on @printable-exercise{true-or-false-abs-val.adoc}.
```
flannery-denny, line 132: Probably makes sense to have students explore `num-abs` before comparing to `negate`

Still not 100% sure it's worth using these named functions if you're not going to have hands on keys.

lessons/Early-Math/absolute-value-and-opposites-2/langs/en-us/index.adoc
```diff
@@ -0,0 +1,135 @@
...
+=== Launch
+
+Explain to students that they are going to receive @printable-exercise{three-sets-cards.adoc, three sets of cards}.
+
+The first set of cards includes Circles of Evaluation that use the `negate` and `num-abs` functions.
+
+The second set of cards translates those expressions into words.
+
+[.lesson-instruction]
+- With your partner, match each Circle card with the corresponding Words card.
```
flannery-denny, line 50: Curious whether Words card is serving us. 
I would describe the words as "Verbal Expressions". At the very least I think it would be valuable to at least sometime refer to them as expressions.

lessons/Early-Math/absolute-value-and-opposites-2/langs/en-us/index.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+[.lesson-instruction]
+- Each row on @printable-exercise{translating-abs-val-opp.adoc} represents a single arithmetic expression, written in three different forms. Sometimes, you will be provided with a Circle to translate, other times you will start with words or a mathematical expression.
+-  Fill in the empty spaces so that all three forms represent the same expression.
+- *MAYBE SHOULD ALSO ADD A WORDS-TO-CIRCLES MATCHING ACTIVITY LIKE IN TRANSLATING LESSON*
+
+=== Synthesize
+
+Here are some key take-aways from this lesson:
+
```
flannery-denny, line 75: Since circles of Evaluation don't exist outside of our curriculum and `num-abs` and `negate` are functions we defined it feels undesirable to have the big idea takeaways focus on those things rather than transferable math understanding... why would a teacher want to do these lessons if they can't see how they connect back to their learning goals?

lessons/Early-Math/absolute-value-and-opposites-2/langs/en-us/index.adoc
```diff
@@ -0,0 +1,135 @@
...
+- You and your partner must agree on each card’s placement before advancing to the next.
+
+Again, circulate during the activity. As students match expressions in words to their arithmetic expressions and Circles, invite them to use the Circle as a tool to help them translate to the other representations.
+
+=== Investigate
+
+[.lesson-instruction]
+- Each row on @printable-exercise{translating-abs-val-opp.adoc} represents a single arithmetic expression, written in three different forms. Sometimes, you will be provided with a Circle to translate, other times you will start with words or a mathematical expression.
+-  Fill in the empty spaces so that all three forms represent the same expression.
+- *MAYBE SHOULD ALSO ADD A WORDS-TO-CIRCLES MATCHING ACTIVITY LIKE IN TRANSLATING LESSON*
```
flannery-denny, line 70: Either tackles this or add to future todos and remove the note from the lesson plan.

lessons/Early-Math/evaluating-exponents/langs/en-us/index.adoc
```diff
@@ -0,0 +1,183 @@
...
+Sudents use Circles, words, and mathematical notation to analyze the structure of expressions with exponents and one operation.
+
+=== Launch
+
+[.lesson-instruction]
+- Show students the following expression: @math{2^3}
+- One way of reading this expression is "two raised to the power of three." Do you know any other ways of translating this expression into words?
+** _Two to the third power; two to the third; two cubed_
+- What does it mean to _cube_ a number?
+** We multiply it by itself three times (literally, compouting the volume of a cube with a side-length of that number)
```
flannery-denny, line 48: I love that you take this there... and... I think if you're going to go there, it deserves a visual of a cube with all 3 dimensions labeled as being 2in or 2cm.

lessons/Early-Math/evaluating-exponents/langs/en-us/index.adoc
```diff
@@ -0,0 +1,183 @@
...
+--
+- Take a look at the expressions and Circles below.
+
+[.embedded, cols="^.^1,^.^1", grid="none", stripes="none" frame="none"]
+|===
+|@math{5^2}							| @math{2^5}
+|@show{(coe  '(num-expt 5 2))}		| @show{(coe  '(num-expt 2 5))}
+|===
+
+- Angela says the Circles are equivalent because of the Commutative Property. Do you agree with Angela?
```
flannery-denny: At this point I would expect to have them create or look at the equivalent circles that don't use the function `num-expt`!

lessons/Early-Math/evaluating-exponents/langs/en-us/index.adoc
```diff
@@ -0,0 +1,183 @@
...
+
+[.lesson-point]
+Exponents are valuable because they act as a shorthand.
+
+The Circle of Evaluation for @math{2 \times 2 \times 2 \times 2 \times 2} looks like this:
+
+[.centered-image]
+@show{(coe '(* (* (* (* 2 2) 2) 2) 2))}
+
+Here is the Circle of Evaluation for @math{2^5}, an equivalent expression:
```
flannery-denny: This section deserves a rewrite. 
I'd suggest consistently connecting expanded form (and its corresponding coe) with exponential form (and its corresponding coe). 

I also think the leap from 2^5 to num-expt(2 5) deserves some more care... and am curious whether the named function is worth introducing in a world without hands on keys...

lessons/Early-Math/evaluating-exponents/langs/en-us/pages/exponents-variables-table.adoc
```diff
@@ -0,0 +1,69 @@
...
+| @showsoln{(coe '(* 5 (num-expt w m)))}
+| Evaluate for @math{w=6} and @math{m=2}.
+
+@ifsoln{@math{5w^m = 180}}
+
+
+
+
+|===
+
```
flannery-denny, line 68: I may have missed something, but I don't think you're asking students to substitute values into the variables in these expressions without ever having introduced them to the concept. In. my classroom, this would definitely require some scaffolding. If I had taught it before I would want to go over an example as a whole class. But I could also imagine not having taught it before teaching a lesson on exponents... which makes me wonder this page should be an `@opt-printable-exercise` at the moment for teachers whose students are prepared for it.

lessons/Early-Math/exponent-expressions-with-negatives/langs/en-us/index.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+=== Overview
+
+Students apply Circles of Evaluation to analyze expressions with exponents and negatives.
+
+=== Launch
+
+[.lesson-instruction]
+--
+- Look closely at each of these expressions, considering the effect of the negation and the exponent on the outcome: @math{-3^2}, @math{(-3)^2}, and @math{-(3^2)}
```
flannery-denny: I'd suggest reordering such that (-3)^2 comes first because it is the most unmistakable example and might lead them to more easily recognize whether the others match the behavior.

lessons/Early-Math/exponent-expressions-with-negatives/langs/en-us/index.adoc
```diff
@@ -0,0 +1,100 @@
...
+- Are any of these three expressions equivalent? How do you know? Draw Circles to help you.
+** _We know that in an expression with no grouping symbols, we evaluate the exponent first - before the negation. Therefore, the first and third expression are equivalent; they both evaluate to @math{-9}. The second expression evaluates to @math{9}. See Circles below._
+[.embedded, cols="^.^1,^.^1", grid="none", stripes="none", frame="none"]
+|===
+|@math{-3^2} or @math{-(3^2)}				| @math{(-3)^2}
+|@show{(coe  '(negate (num-expt 3 2)))}		| @show{(coe  '(num-expt (negate 3) 2) )}
+|===
```
flannery-denny, line 50: I'm not sure how the circles would help students to know that -3^2 and -(3^2)... but I do think the visual draws a powerful distinction from (-3)^2. That makes me think we want every student to see these circles, so perhaps they belong on a workbook page.

Wondering if you might instead want to show coe for -(3^2) and (-3)^2, have them describe how they're different. Then tell them that -3^2 has the same circle as one of the two and have them try to decide which and justify their thinking.

lessons/Early-Math/exponent-expressions-with-negatives/langs/en-us/pages/true-or-false.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+@showsoln{(coe '(negate (num-expt 2 6)))}
+@ifsoln{@math{=}}
+@showsoln{(coe '(num-expt (negate 2) 6))}
+| @ifsoln{False: @math{-64 +≠+ 64}}
+
+
+|===
+
+@n What do you notice about the Circles on the _left_? @fitb{}{@ifsoln{On the left, the _outer_ circle is always negating - so these Circles always have a negative}}
```
flannery-denny, line 75: I'm a little uncomfortable with this explanation because negate(negate(3)) would have negate on the outside and evaluate to a positive... and that's definitely a problem that my sixth graders had to confront.

flannery-denny (changes requested):
As discussed in our call, I'd suggest a rewrite of the Associative Property Lesson:
- [x] The Associative Property lesson would be much more powerful if you focused on the power of regrouping rather than the mechanics of its possibility. 2 X 17 X 5 can be done in lots of orders... but if a student notices that they could group 2 x 5 = 10 it becomes a simple mental math problem... and, to me, this is the essence of why a student should care about the property!
- [x] My understanding of the associative property is fairly aligned with this definition
> Associative property states that the sum or product of three or more numbers remains the same, irrespective of how they are grouped
- [x] The current distinction between which problems can be solved with the Associative property  and which require both associative and commutative feels narrow... and leaves kids (and teachers!) focused on definitions rather than giving them permission to think flexibly.
- [ ] I'm not sure circles of evaluation do as good of a job of making this regrouping capacity visible as other connector symbols... which makes me think we may have to also show other representations and address that concern (are we regrouping or reordering?) explicitly. 
- [ ] my most concrete suggestion is to start the problems with expressions rather than circles of evaluation. Have students draw all possible coe. And then make sure to ask them to reflect on which makes their work most efficient.
- [ ] might be worth discussing in our next curriculum meeting?

Absolute Value & Opposites:
- [ ] I am not totally convinced by these lessons... the card sort (perhaps there should also be a Desmos version of this?) and matching circles and expressions page (presented kind of as an after thought) feel most compelling...
- [x] I have concerns about the dominant role that the `negate` and `num-abs` functions are playing in these lessons given that they don't exist outside of our materials. Want to make sure we a) are providing enough support in how we introduce the named functions in this pathway b) are making sure to transfer understanding they are generating back to universally communicable math understanding.

Overarchingly:

- [ ] As discussed in previous pull request, "Circles" doesn't have any concrete meaning related to this work. I have pushed a lot of changes to reword to "Circles of Evaluation". I expect I missed some and that you will keep writing things. Please orient toward using the full phrase.
- [x] In general students this age are unlikely to have experience with the word functions and it probably makes more use to talk about operations or operators (which aligns with Pyret).
- [ ] You have a lot of student-facing directions on workbook pages asking them to analyze things. I am not sure what I am supposed to do to satisfy that direction. My guess is that, while analyze might be something a teacher suggests students do, a workbook page should never use analyze in the directions, opting for more concrete deliverables instead such as evaluate.
- [x] I have made some changes to get rid of the word "worksheet". We don't refer to any of our materials as worksheets. Use "workbook page" whenever necessary to generalize.
- [ ] Figure out what directions you want to use on workbook pages instead of "Analyze" and deploy throughout.
- [x] Deploy `(?ANS)` throughout workbook pages.
- [ ] Scale back any worksheet that is currently more advanced to focus on whole numbers and make an issue adding intentions for plans to add more advanced material in the future
- [ ] Some of these lesson never refer to the word expression and only refer to Circles of Evaluation. I think it would behoove us to make sure that we are thinking COE as a tool rather than presenting it as the sole representation of mathematical expressions.
- [x] I am a little uncertain about our use of the word equivalence. I think of 5(2+5) as being equivalent to 5x7 and 10 + 25 and 35, but not to 30 + 5, although it evaluates to the same thing. I am not sure where the canon lands on that.
- [ ] Noting that this pathway does not make much use of Notice & Wonder pedagogy and could.
- [x] FYI - A number of workbook pages are currently 2 pages long - **RT needs @schanzer's help to fix the one that remains.** 

> WARNING: in early-math workbook-long-sols.pdf, ../../lessons/identity-property/solution-pages/discover-identity.pdf has 2 pages

lessons/Early-Math/absolute-value-and-opposites-2/langs/en-us/index.adoc
schanzer: I LOVE the idea of "verbal expressions". 100% agree with @flannery-denny that keeping the word "expression" is key

lessons/Early-Math/identity-property/langs/en-us/pages/discover-identity.adoc
retabak: Flannery, you're so good!

lessons/Early-Math/identity-property/langs/en-us/pages/which-one-doesnt-belong-variables.adoc
retabak: the first one doesn't.

schanzer:
@retabak I fully agree with Flannery's "analyze" comment. I don't think it's useful in the way you are using it, and I suspect a decent fraction of those could be replaced with "notice and wonder"

retabak:
I agree, too! Haven't resolved the comment because it's all over the place.

PR #1244, opened 2023-01-09
revised stress or chill lesson

retabak:
I updated "Stress or Chill" and would love feedback!

A few changes I made:
- I organized the questions into a few different groups to highlight the different thing this project accomplishes.
- I restructured this to match up with the Data Cycle.
- I provided the Google survey (rather than having kids/ teachers build it themsleves) to create less busy work.
- I made it MUCH clearer that kids are looking at classwide data.

I still need to make a rubric, but wanted to get feedback on these changes first.

Question to conisder: Right now, this project is in "box plots" and I have absolutely no clue why it is there. Where do y'all think would be a better place for this? Or, can you explain to me why it's in "box plots"?

Another question! Joy had students use a random number generator to decide when to collect data. It's unclear if she wants all students collecting data at the same two times each day, or if each student should generate their own unique numbers. It's also unclear if the numbers should include, like, 3am, or if those should be excluded. I think the most interesting project would result from excluding sleeping hours, and from every student collecting data at different times.... but I could see arguments for other options.  What do you think is best here?

@flannery-denny - not sure how much is on your plate or if you'll have time to take a look but I've included you as a reviewer anyhow.

### DIFF
--- a/lessons/box-plots/langs/en-us/pages/rubric-stress-or-chill.adoc
+++ b/lessons/box-plots/langs/en-us/pages/rubric-stress-or-chill.adoc
@@ -1,4 +1,4 @@
-= Food Habits Rubric
+= Stress or Chill? Rubric
 
 [cols="2,4,4,4", options="header"]
 |===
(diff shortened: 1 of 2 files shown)


schanzer (changes requested):
@retabak this is MASSIVELY better!! I have only a few small questions/proposed changes, but I think this is very close to being mergable

retabak:
One last thought -- for teachers who want a smaller project, I wonder if it would be helpful to suggest that they could completely remove the personality assessment piece (which I actually don't think adds all that much? But I haven't done the project). 

 Or, we could just remove the personality assesment piece altogether, on the grounds that it makes a big project even bigger and more ambitious. The personality stuff also feels a little like an afterthought, in my opinion. (The personality assessment is of great importance in the original IDS lesson, where this project is all about merging two seperate datasets.)

schanzer (approved):

PR #1300, opened 2023-03-23
multiplicative-inverse and additive-inverse

retabak:
Responded to your feedback (in multiplicative inverse lesson), then created additive-inverse lesson. Take a look!

Need to add glossary definitions still.


lessons/Early-Math/multiplicative-inverse/langs/en-us/pages/multiplicative-inverse.adoc
```diff
@@ -0,0 +1,83 @@
...
+td > .content > .paragraph > * { vertical-align: middle; }
+</style>
+++++
+
+On the line next to each value, give the multiplicative inverse.
+
+[.FillVerticalSpace, cols="^.^3,^.^3,^.^3,^.^3,^.^3", stripes="none"]
+|===
+
+| @math{\frac{4}{5} } @fitb{5em}
```
schanzer: Why not make these look like the table below, and even _fuse_ the two top tables entirely? Everything should just be `A * B = 1`.

schanzer (changes requested):
These look awesome! A few small changes suggested, mostly to workbook pages.

schanzer:
@retabak Git isn't letting me add new comments (apparently because the old ones are still open?), so here they are:

- In discover-inverse.html, I think you need to show students at least one more circle in order to have them notice&wonder about a pattern
- I _love_ subtract-first-left-to-right, and divide-first-left-to-right!
- The multiplicative-inverse.html file has the same issues as the additive-inverse one: use a consistent format for the whole page (this is in my earlier review, and hasn't yet been addressed)
- I really, really love the way you've given these lessons an almost identical structure. That makes it way easier for teachers to implement, and for students to tackle. You might want to put this in the notes for the second lesson, recommending that students first look back at prior pages. (Note that if you do this, you'll need to add additive-inverse as a prerequisite)

Overall I think you're very close to being done, but you still have a bunch of unaddressed comments from March

schanzer (changes requested):
See comments from the review I began in March

Issue #1302, opened 2023-03-25
Address Flannery's comments on Algebra workbook


flannery-denny:
I'm attaching pdfs, which should amount to scanned copies of all of the workbook pages with my annotations. I'm assuming @retabak has already caught much of this and can handle most of it, but @schanzer, it feels like some of this would be best resolved through css, so please give the notes a look and see if there's anything it makes more sense for you to tackle.

Noting that some of my suggestions about clarifying directions about what pages can and can't be tested in cpo is in response to teacher questions on discourse.

A few overarching things to note.
- [ ] the notes pages feel crowded and it may be more efficient to write a css rule that could be added to most notes pages to give a little more space between the bulleted items so that they are easier to read.
- [ ]  the hs look tiny in what I remember as being function names
- [ ] there are still some instances where we are showing too many parentheses in the circle of evaluation pages
- [ ] I have an overarching question for the order of operation pages re: whether we want to use the phrase "arithmetic expression rather than "math expression"
- [ ] I'm noticing that we've standardized how we format solutions that get written out (hooray) but haven't standardized ho we identify solutions when we're trying to identifu which option they should have circled and sometimes it's hard to recognize that the formatting for these selections is different (eg. how functions fail the vertical line test)
- [ ] @schanzer I'm pretty sure I found some serious mistakes in our sample responses in the distance lesson. Please doublecheck that my suggested fixes are correct as my head has been out of the game for a while.
[OCD-CODOFIL_20230324_181249.pdf](https://github.com/bootstrapworld/curriculum/files/11067512/OCD-CODOFIL_20230324_181249.pdf)
[OCD-CODOFIL_20230324_181645.pdf](https://github.com/bootstrapworld/curriculum/files/11067513/OCD-CODOFIL_20230324_181645.pdf)
[OCD-CODOFIL_20230324_181956.pdf](https://github.com/bootstrapworld/curriculum/files/11067514/OCD-CODOFIL_20230324_181956.pdf)
[OCD-CODOFIL_20230324_182252.pdf](https://github.com/bootstrapworld/curriculum/files/11067515/OCD-CODOFIL_20230324_182252.pdf)

PR #1308, opened 2023-03-30
Algebra workbook review

retabak:
Addresses Flannery's concerns from #1302 along with a slew of other issues!


schanzer (changes requested):
Two small nits. Otherwise looking good!

Issue #1294, opened 2023-03-13
square roots look funny in printed workbook


retabak:
... but fine in workbook PDFs!

Images are from the first item on translate-arithmetic-to-circles-and-code-challenge.adoc from Order of Operations.

![image](https://user-images.githubusercontent.com/87144723/224831500-73e5b2d9-cb64-42c2-8fed-c02bb2527a17.png)

![image](https://user-images.githubusercontent.com/87144723/224831516-48c991d7-7408-4e76-b40b-e7d67c0fcf9c.png)

retabak:
Huh! I think that sounds like a good move.

retabak:
Adding a note that we need to also check the letter h in code font, which appeared teeny tiny on the printed version that we reviewed.

retabak:
Funny, I kinda thought the distance formula looked okay, without the bar extending to the right. I was more bothered by tops of numbers getting chopped off!

I tried the live demo thing and my h looked okay?
![image](https://user-images.githubusercontent.com/87144723/230971691-4fe03347-54d8-4da8-8264-014efcf3a9f8.png)

schanzer:
@retabak have you looked at the distance formula? It's nonsense without the bar extending.

In your screnshoot, the `g` looks small to me.

retabak:
Yeah, but h is the only weird letter on the workbook that I reviewed?!

![image](https://user-images.githubusercontent.com/87144723/230973526-22d0ca73-7bd7-4d3f-9dc9-7baebdc4b34e.png)


Also, isn't this an acceptable way to depict a radical symbol? Without a bar extending? If I google "radical symbol" this is what I get: √ 

![image](https://user-images.githubusercontent.com/87144723/230973685-61f92e33-3fcc-46da-8e30-5737c0338162.png)

retabak:
(That said, it'd be cool if the bar extended! Probably better, even! But I think this way is.... fine.)

schanzer:
This is now fixed in #1318. I'll leave this issue open until the PR is merged.

PR #1318, opened 2023-04-10
MathJax3

schanzer:
This PR moves us from v2 to v3, and is designed to address a number of issues:

- The "roof of the house" in the Square Root symbol displaying/printing incorrectly in certain cases (see #1294)
- Lettering is inconsistently sized (see #1294)
- Clean up some technical debt from using an ancient library
- Improve page load times - and therefore PDF generation speed! (MathJax3 is more than 2x as fast, which shaves ~100ms off each page)
- Reduce the size of our JS dependencies (MathJax3 is modular and can be tree-shaken)


retabak:
I did a quick once-over. Here are some thoughts:

- [ ]  Why is the "roof of the house" on the square root symbol sometimes (not always) so dark now?! I suppose it's better than before! (This is page 17 of workbook-long-sols; it appears normal, however, on page 32.)

image

- [ ]  Arithmetic Exp to COE and Code - challenge 2 needs wider first column
- [ ] **SOMETHING WEIRD happening on p. 75**

image

- [ ]  Just noticed that the solutions for "Welcome to Alice's Restaurant" are all wescheme?!? Has that always been the case?
- [ ]  Challenge: Character Movement in Two Directions -Q1 needs `update danger` to be hyphenated.
- [ ] Oooh, those radical signs on page 107 look GREAT!
- [ ] UGGHHHHH just noticed the Circle of Eval on page 108 doesn't have lins separating the function from the rest of the circle. 
- [ ] Last page of contracts has several lines with two # rather than one. (I corrected this for the linse that have contracts, but not for the blank lines.)

schanzer:
@ds26gte looks like Rachel found a bug related to mathjax generation. In `function-definition-linear/pages/definitions-from-tables.adoc`, we're getting raw HTML output like this:

`${empty}$\displaystyle f(x) =${empty}$`

This looks like a possible preproc issue. Can you investigate?

schanzer:
@retabak you wrote " Arithmetic Exp to COE and Code - challenge 2 needs wider first column". Why? Does one of the lines wrap for you?

ds26gte:
> @ds26gte looks like Rachel found a bug related to mathjax generation. In `function-definition-linear/pages/definitions-from-tables.adoc`, we're getting raw HTML output like this:
> 
> `${empty}$\displaystyle f(x) =${empty}$`
> 
> This looks like a possible preproc issue. Can you investigate?

Looks like generating `${empty}$` in the `.asc` file with the expectation that it will get converted to simple `$$` in the HTML doesn't always work. I've reverted to generating a marker and then converting it during postproc to `$$`. However, the marker is simpler than the previous 3-part markers, since we only need something to stand in for `$$`.  

Fix in commit ad929b91b in branch MathJax3.

PR #1322, opened 2023-04-12
more Codap !

retabak:
Okay - I've applied your suggested changes to codap-bar-charts, and also continued to improve codap-exploring (just minor adjustments).

I've started restructuring the instruction around transformers in a major way....

- codap-transformers1 introduces filtering. 
- codap-transformers2 introduces sorting. I'd like to add a section on composing table operations to the end of this lesson, which has students composing with just filter and sort. (This was a suggestion from the Richmond notes.) Does that still feel like a good move to you?
- codap-transformers3 will cover transform and build (but I haven't revisited this lesson yet).
- next would come composing-table-operations but with all build, filter, sort and transform.

I'm still concerned at how contract + purpose statement are kinda just squished into the lesson on filtering. Was wondering if there are any workbook pages that already exist that could be added in as optional practice? Or, do you have any other ideas for organically adding that content but not overwhelming students / teachers?


lessons/codap-transformers2/langs/en-us/index.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+[.lesson-instruction]
+- When manipulating data, what do you think might be some advantages and disadvantages to changing the actual table?
+** _Advantage: It's quick, easy and intuitive._
+** _Disadvantage: If I want to refer to the previous arrangement of animals for any reason... it might no longer exist._
+- What are some advantages and disadvantages to creating a new modified table?
+** _Advantage: I will be able to refer back to my old table, and maybe even compare the new table with the old table. I can use my Transformer again later once I save it!_
+** _Disadvantage: It will take a bit of typing to set it up, and a few extra clicks._
+
+You will direct students to sort various columns on the table by clicking the attribute name and then selecting from the drop-down menu either `Sort Ascending` or `Sort Descending`. This will modify the original table.
```
schanzer, line 48: You're assuming students know what "ascending" and "descending" mean. Is that a safe assumption? Is it worth explaining, just in case?

lessons/codap-transformers2/langs/en-us/index.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+[.lesson-instruction]
+- Let's sort the animals alphabetically by name and see if we get the same answer as we did before! _(The image below shows where to click to sort the animals alphabetically by name.)_
```
schanzer, line 51: alphabetically = ascending, yes? This is why you might need to define ascending and descending, above

lessons/codap-transformers2/langs/en-us/index.adoc
```diff
@@ -0,0 +1,81 @@
...
+=== Investigate
+
+Now, let's try sorting with a Transformer!
+
+[.lesson-instruction]
+Complete the @printable-exercise{pages/codap-sort-transformer.adoc} worksheet, which will walk you through creating, applying, and saving a `Sort` Transformer.
+
+@right{@image{images/sort-transformer.png,250}}
+
+Initially, students might question the value of using a Transformer to sort. That is okay! The sort functionality embedded into CODAP is a more efficient choice only in some contexts.
```
schanzer, line 66: I think this is more of a synthesize question OR workbook page question. We should ask students to think about what those contexts might be!

lessons/codap-transformers2/langs/en-us/index.adoc
```diff
@@ -0,0 +1,81 @@
...
+- Can you sort the animals by age, from youngest to oldest?
+- How about from heaviest to lightest?
+- Now sort the animals table by how long it took for each animal to be adopted, in ascending order.
+
+=== Investigate
+
+Now, let's try sorting with a Transformer!
+
+[.lesson-instruction]
+Complete the @printable-exercise{pages/codap-sort-transformer.adoc} worksheet, which will walk you through creating, applying, and saving a `Sort` Transformer.
```
schanzer, line 62: The title of this page is "Sort Transformer", but in the previous lesson the transformer page is just "Filter" and in the next lesson you have "Transform" and "Build Attribute".

These titles should be standardized, and **I think these pages would really benefit from having a shared structure that introduces each one.** The last two (Transform/Build) have an identical structure, and Sort isn't that far off. Filter is the forgotten cousin here, and could use some love to bring it back into the family

lessons/codap-transformers2/langs/en-us/pages/codap-sort-transformer.adoc
```diff
@@ -6,29 +6,31 @@
...
-- We want to give each Transformer a useful name, one which clearly describes what that Transformer does. Generally, a brief hyphenated description works - for instance: `order-by-pounds`.
+@n We're going to create a Transformer to sort the animals alphabetically by name. Select Sort from the drop-down menu.
 
-@n What would be a good name for a Transformer that organizes the table alphabetically by animals' names?
+@n What would be a good name for a Transformer that organizes the table alphabetically by animals' names? @fitb{}{@ifsoln{order-by-name.}}
 
-@fitb{}{@ifsoln{order-by-name.}}
-
-- Type in the name you wrote above.
+_Type in that name._
```
schanzer, line 19: Where? If you're going to take up a whole line of text to give explicit instructions, let's use the whole line and be super-clear. Heck, maybe a screenshot of a transformer would be helpful on this page, with the fields numbered? Then you could say "type that name into box 1". 

Then rinse and repeat for the other transformer pages

lessons/codap-transformers2/langs/en-us/pages/notes-sort-transformer.adoc
```diff
@@ -0,0 +1,14 @@
+= Using Transformers: Sort, Filter, Transform Attribute, Build Attribute
```
schanzer, line 1: Shouldn't this be in codap-transformers1?

lessons/codap-transformers2/langs/en-us/pages/notes-sort-transformer.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+*Transformers* allow us to transform datasets to produce new, distinct output datasets, instead of modifying the original input dataset itself. We use them to manipulate tables and enable low-stakes "what if?" exploration.
+
+- In this course, we'll be using a variety of Transformers.
+
+** `Sort` - We must provide `Sort` with a dataset, a sort method (we'll focus on sorting by attribute in this course), and a sort direction. `Sort` produces a copy of the input dataset, with cases sorted by the selected attribute.
```
schanzer, line 7: I'd re-order these to use the same order in which they are introduced

lessons/codap-transformers3/langs/en-us/index.adoc
```diff
@@ -0,0 +1,74 @@
...
+- A dataset from Europe might list everything in metric (centimeters, kilograms, etc), so we use `Transform Attribute` to convert to imperial units (inches, pounds, etc).
+- A dataset about schools might include columns for how many students are in the school and how many of those students identify as multi-racial. But when comparing schools of different sizes, what we really want is a column showing what _percentage_ of students identify as multi-racial. We could use `Transform Attribute` to compute that for every row in the table.
+
+== Building Columns @duration{10 minutes}
+
+=== Overview
+Students learn how to _build columns_, using the `Build Attribute` Transformer.
+
+=== Launch
+So far, we've used Transformers to _filter_ and to _transform an attribute_. The final Transformer we are exploring today is called `Build Attribute`. Can you guess what this Transformer does and how it might be similar to _and_ different from `Transform Attribute`?
```
schanzer, line 53: I think this "can you guess" is too hard for kids. How _could_ they guess? Instead, maybe give an example of where you want the transformed column but you DON'T want to lose the original. Present the problem that is solved by Build Attribute

schanzer (changes requested):
Overall I think this is a major improvement. Bravo!

I've checked in a few wording changes and fixes, so you should start by reviewing them and making sure they're ok with you.

The single biggest piece of work in front of you is to improve those "Introducing Transformer X" worksheets so they all have a common look and feel. You have a TON of room on the last 2-3 for a helpful screenshot, and I bet you can find room on the first one as well.

As for your question about Contract+Purpose: I think you could copy one of the matching activities from `functions-dr`, and replace it with table-specific functions. Can students identify which contract/purpose pairs match which word problems?

I think you do a decent job introducing them in the Filter lesson, but you never really say much about purposes after that lesson at all! Find a consistent way to spiral back to them as students add new transformers. Perhaps a matching page in each lesson, with the matches starting out as only filters but gradually having more and more used on the same page? This would help students practice differentiating between Transformers, and drive home the importance of purpose statements throughout

lessons/codap-transformers2/langs/en-us/pages/notes-sort-transformer.adoc
retabak: yeeah, things got shuffled and need a bit of reorganization.

lessons/codap-transformers2/langs/en-us/pages/notes-sort-transformer.adoc
retabak: good call.

lessons/codap-transformers3/langs/en-us/index.adoc
retabak: also a good call!

PR #1506, opened 2023-08-30
Autogen Slides

schanzer:
@ds26gte I've created a branch (`autogen-slides`) that uses the `scatter-plots` lesson as a sandbox. I chose this lesson because it's got the 2nd-largest slide deck in our library, and gets an order of magnitude more use and polish than the 
largest one (standard-deviation).

Below are my notes about slide parsing. 

### Course String
- this is determined based on the folder hierarchy of the lesson folder
- Data Science and Algebra 2 -> "DS"
- Algebra and Early Math -> "A"
- Everything else -> ???? How about "Default" for now?

### Title Slide and Metadata
- title should be extracted from the lesson title
- layout should be extracted from the course string
- contents should be pretty boilerplate:
```
---
{layout="<Course String> Title Slide"}
# <lesson title>

<!--
To learn more about how to use PearDeck, and how to view the embedded links on these slides without going into present mode visit https://help.peardeck.com/en

<metadata -- some reasonable summary of prerequisites, standards, materials, facilitator notes, etc. Could simply be a series of lists with links>
-->
```

### Identifying Slide Breaks and Layouts
- [x] A slide boundary occurs at the start of any 2nd- or 3rd-level header (`==` or `===`), or immediately after the made-up atomic directive `@slidebreak`
- [x] Layouts are of the form `<Section><Image-Orientation>-<Suffix>`. 
- [x] <Section> is mandatory and comes from:
- [x] 2nd-level headers  `<section>` (e.g. - Launch, Investigate, or Synthesize) - (Ignore Overview sections)
- [x] the 1st-level header "Additional Exercises" found at the end of some lesson plans - which translate to `Supplemental` slide layout and should contain everything from that lesson plan section.
- [x] `<Image-Orientation>` is optional and comes from:
-- If no image appears within a slide, the image-orientation is empty
-- If only one image appears within a slide, the Image-Orientation should always push images to the _right_, triggering the layout being LaunchR, InvestigateR, etc. One exception: `@center` should trigger LaunchC, InvestigateC, etc. If no enclosing directive exists, default to R. 
-- If more than one image appears, the slide layout should be determined strictly from the most-recent 2nd-level header, and positioning directives should be ignored -- authors will have to manually position the images relative to one another
- [ ] If the slide contents include `@lesson-instruction` and the proglang is anything aside from `none`, the layout should have `-DN` appended to it. (e.g. - Investigate-DN, InvestigateR-DN, etc)
- [ ] If the slide contents include `@lesson-roleplay` then the suffix is RP

_It may very well turn out that there's no reliable way to infer anything beyond the Section, but this would be a pretty substantial labor-savings_

### Slide Contents
- [ ]  `@show{(contract...)}` should be handled the way `@show{(coe...)}` is: fully converted into HTML, then rendered with `$$$dom...$$$`
- [ ]  If a "Common Misconceptions" section exists (3rd level), the entire contents of the section should be prepended to the _first slide of the Synthesize section that follows_. (A "common misconceptions" section that is not _immediately followed by a synthesize section_ should be a static error)
- [x] Title should be extracted from the most-recent 2nd-level. For example, in a lesson section called "Foo" (`== Foo`), all titles should be "Foo", regardless of layout
- [x]  `@teacher{...}` and `@strategy{....}` contents will populate the slide notes (`<!-- .... -->`). If more than one `@teacher` section exists within a slide, they should be appended to one another with a few line breaks to separate them
   -> When `@span{.title}{XYZ}` is found within a @strategy box, it should be translated as `**XYZ**{style="font-size:12pt}`
- [x]  Q&A should be translated according to #1518 
- [x]  `@lesson-point{XYZ}` should become `**XYZ**{style="font-size: 22pt}`

### TBD
- Really long bullets that refer to a common image. It would be nice to find a way of auto-detecting them, and splitting them across multiple slides with the same layout and image. This is probably beyond automation for the foreseeable future.


ds26gte:
Could you add a spec for how to determine from the header what the layout is to be (Launch, Investigate, or Synthesize)?

schanzer:
@ds26gte the layout will always start with Title, Launch, Investigate, Synthesize, or Supplemental.

- Title is always the first (and only) slide
- Supplemental is always the last slide
- Everything else is determined by the most-recently-seen second-level heading

schanzer:
@ds26gte a few errors:

-  I'm still not seeing @q-and-a handled correctly.
- lesson-point is sandwiching the lesson-point contents between double-asterisks (`**`), such that the asterisks are on their own line. They need to be on the same line as the content

schanzer:
@ds26gte 

- @QandA is putting the correct content on the slides, but the mix of questions-and-answers are missing from the slide notes
- I'm still seeing the same problems with lesson-point. [Here's a slide deck](https://docs.google.com/presentation/d/12JTa7A5Flhz8tKqr6k_T1cbgPWlmatZmlzvo-VBXJP8/edit#slide=id.fed24362-5f27-11ee-8829-850e2663a4d1) from the most recent pull

schanzer:
@ds26gte this is really taking shape nicely! One addition to the spec (already added above):

When `@span{.title}{XYZ}` is found within a @strategy box, it should be translated as `**XYZ**{style="font-size:12pt}`

Issue #1500, opened 2023-08-28
create rubrics for all early math projects in PPS scope and sequence and SOME project pages


retabak:
Alignment guides: [Grade 6](https://docs.google.com/document/d/17xVBukZ55njPAQNVYWaI4DQviQN_LbdvuDF3et-dbO4/edit?usp=sharing), [Grade 7](https://docs.google.com/document/d/1LcsXkNC-IaRbjwPI9z3wyewv56H9LRMo-OACpmNMej8/edit?usp=sharing), and [Grade 8](https://docs.google.com/document/d/11QGTMlxymgB8hGUGcNlLUa9crsChXSPH3tbb_Z8MSbM/edit?usp=sharing)

**Grade 8:**

- [x] Unit 2 Project (~2 days): Students come up with a function of their own, which takes a shape and manipulates it using at least three transformations. They submit their Pyret files, which contain the function definition.
- [x] Note that units 3&4 have no projects. Could there be something?

**Grade 7:**

- [x] Unit 1 Project (~5 days): Making a logo using the shape producing functions, along with rotate, scale, overlay, above, below, and beside. As part of their projects, students should submit: A sketch of the logo they want to draw; A Circle of Evaluation, showing how their functions are composed to make the logo; A working program that generates that logo
- [x] Note that units 3&4 have no projects

**Grade 6:**

- [x] Unit 2 Project: Print, and construct a rectangular prism - how many different nets can you come up with to create the same shape?
- [x] Unit 3 Project: Have students complete [this workbook page](https://bootstrapworld.org/materials/fall2023/en-us/lessons/evaluating-exponents/pages/exponents-variables-table.html), then give them the solutions for the third column. Have them type the code in Pyret and see if they were right. Have teams of students create challenges for one another, using opposites, absolute values, and exponents.

schanzer:
@retabak how is this different from #1495 ?

retabak:
Lol. It is exactly the same. I forgot that I made that. This one is clearly more detailed. Gonna close the old one.

retabak:
Identify projects that will be taught before December.

@schanzer gives me the thumbs up on creating project pages for projects to be taught before December.

retabak:
started doing some thinking here: https://docs.google.com/document/d/1cMjRb13AUW5tcql6x4fm47LSCk7skRbICWzTq-bB-WE/edit

PR #1546, opened 2023-09-29
logos project first draft! (#1500)

retabak:
I am very proud of this and how quickly it came together!

I'm not sure if all pages need to be @opt-printable so that this can go to the web? Maybe it won't matter since they're in the early math pathway which doesn't have printed workbooks at the moment. LMK.

also, I've included work from Joy's students as samples. Should probably find a way to credit them / check in with Joy to see if it's okay.

This project + rubric was the most urgent because it is in the first unit. The other two projects are in unit 2. Hoping to finish those up tomorrow or Monday.


schanzer (changes requested):
@retabak this is a really nice first draft! I made a number of changes to the formatting and wording -- please make sure you're okay with them.

I think there are four more significant changes that need to happen before this is done:

- [x] The sample file is kind of unreadable. We need to model proper line-breaking and indenting for students, which should also be called out in a workbook page so students see what it's for. (We should probably do the same in `flags` as well!)
- [x] When you ask students to annotate their shapes (listing colors, shapes, transformations...), why not have them fill out a table similar to the one we use to scaffold this same process in flags?
- [ ] Without providing scaffolding for a student to review _another_ student's work, you run the risk of kids giving little constructive feedback. I'd suggest making another page, which has separate questions for "what do you like about this logo?" and "what suggestions do you have to make it even better?" You could probably do this using 2/3rds of the page, and then move the "after getting feedback from a friend..." question from the create-your-logo page to the bottom third of the new one.
- [x] Finally, and this is the big one -- why are you having students type their logos into the interactions area instead of the definitions area? This is a recipe for frustration, since the code will almost certainly be more than one line long, and a kid who uses the up-arrow will immediately delete all of their code. Multi-line code belongs in the definitions area. This also resolves the acrobatics you're doing to try and help them move the code from one place to another.

retabak:
@schanzer! Ready for you to take another look. Feeling good about this.

schanzer:
@retabak this looks way better! A few small nits:

- [ ] The first lesson-instruction should really be a @QandA. I would change this myself, but I think it's better for you to get some more practice with it
- [ ] In Sample Logo Study, are you sure "indentations" is a word providence 6th graders will understand? See my proposed changes
- [ ] In Sample Logo Study question 2, you write: "Discuss with your class why the code is formatted that way, then follow the directions below" I think you need to cut out the whole sentence. You don't really expect the kids to have a whole-class discussion on their own, right?
- [ ] I really don't like the phrase "draw the logo of your dreams". This feels simultaneously way too vague, way too high pressure (what if I don't dream about logos, or I'm not artistic?), and like setting kids up for disappointment when their artistic ability outstrips their programming ability. Can we use some different wording here? Or maybe even some explicit instruction in the lesson, like "notice how almost all major logos are composed of simple shapes, bold solid colors, etc"
- [ ] In Code Your Logo, you have some extra text at the end that seems teacher-facing. I've pushed an edit. See what you think.

PR #1606, opened 2023-10-26
threats to validity project draft

retabak:


lessons/Projects/project-threats/langs/en-us/index.adoc
```diff
@@ -0,0 +1,92 @@
...
+Congratulations! You're a data scientist who has been granted funds to conduct research on a topic of your choosing! There's one complication...
+
+You're a *BAD* data scientist!
+
+For this project, you and a partner will develop a statistical question that you would be interested in exploring. It must be, however, a statistical question that lends itself to many threats needing to be addressed.
+
+@teacher{Note: We recommend printing and distributing the student-facing @handout{rubric-threats-to-validity.adoc} to help students understand the scope of the project - and your expectations - at the outset. Teachers are welcome and encouraged to edit and adapt the rubric for their unique classroom context.}
+
+@lesson-instruction{
+- Review with your partner the four threats to validity to ensure that the two of you understand what each one entails.
```
schanzer, line 55: "Review with your partner the four threats..." You have this really consistent, really fascinating grammatical preference that puts the subjects before the noun. I see it over and over, and if I didn't know you I would SWEAR you're someone who learned english after first learning a different language, and are mapping it on top of your native grammar.

It's not _wrong_ at all - just highly unusual, and more complex of a read.

Anyway, I would re-word it as "Review the four threats with your partner...". Ditto for every other place you use this. Instead of _Verb, Subject, Noun_ you want _Verb, Noun, Subject_.

lessons/Projects/project-threats/langs/en-us/pages/threats-project-notes-2.adoc
```diff
@@ -0,0 +1,28 @@
+[.landscape]
+= Analysis of Research Plan Predicted Outcomes (2)
+
+Complete the table below with your partner.
+
+@vspace{1ex}
+
+[.FillVerticalSpace, cols="^.^3,^.^5,^.^5", options="header", stripes="none"]
```
schanzer, line 8: It really seems like the first column should be even smaller. Also, all this empty space feels like we're begging for kids to write as messily as possible.  Can we either fill these cells with blank lines, or replace the table with numbered-questions followed by blank lines?

(Ditto for threats-project-notes-2.adoc)

lessons/Projects/project-threats/langs/en-us/pages/rubric-threats-to-validity.adoc
```diff
@@ -0,0 +1,48 @@
+= Threats to Validity Rubric
+
+[cols="2,4,4,4", options="header"]
+|===
+|
+| Wow!
+| Getting There
+| Needs Improvement
+
+| Statistical question
```
schanzer, line 10: I think I need to hear move about why the question itself belongs here. Somehow I'm just not able to wrap my head around choosing a hi-threat v. lo-threat question! But obviously you do, so I want to hear it from you.

schanzer (changes requested):
Mostly small comments throughout the PR. One major missing piece -- what about when students trade papers and try to "spot the slime"? I would think there would need to be one or more pages to scaffold students as they search through their friends' "bad" projects.

lessons/Projects/project-threats/langs/en-us/index.adoc
retabak: no one has pointed this out for me ever before! It really truly doesn't feel any more complex to me!? Regardless, I see that it bothers you, so I will fix it and stay on the lookout for it in the future! 

That said, the phrase "with your partner" feels important to me, so I didn't want it to be lost at the end of the sentence. "Reviewing with your partner" is very different from "reviewing." I suppose I could say, "With your partner, review the four threats..." That would satisfy everyone.

(In case you care - your terminology is off. I was using "verb - prepositional phrase - predicate," but you want "verb - predicate - prepositional phrase.")

lessons/Projects/project-threats/langs/en-us/pages/rubric-threats-to-validity.adoc
retabak: nah, I agree with you. It can go!

lessons/Projects/project-threats/langs/en-us/index.adoc
schanzer: You're absolutely right. I want predicate before prepositional phrase. :) 

But yeah - you never see the other phrasing in American children's books. Actually, now I'm wondering if it's a British-English thing vs. an American-English thing....

retabak:
"spotting slime" wasn't a part of the original project. I think you're thinking about "design a survey" in the collecting data lesson. (Turns out there are a bunch of data science lessons that weren't on my list.... that was one.) That said, I think it would be a great idea to add it, and that I could do so relatively easily.

PR #1638, opened 2023-11-14
Materials cleanup

schanzer:
This PR completes our overhaul of the materials section:

- Instead of listing every page in the summary table, we show a single link to download all pages as a (dynamically-generated) PDF
- This link is followed by a small button, which will show/hide the list

Before:
image

After:
image 
image


lib/preproc.rkt
```diff
@@ -587,9 +587,8 @@
                (let* ([styled-link-output
                         (let ([g-pdf (path-replace-extension g ".pdf")]
                               [tack-on ", window=\"_blank\""])
-                          (format "[.PrintableExercise]##~a {startsb} link:~alessons/pass:[~a][html~a] &#x7c; link:~alessons/pass:[~a][pdf~a] {endsb}##" (or page-title link-text)
```
ds26gte, line 590: Since you're not using g-pdf anymore, its introduction on line 588 can go

ds26gte (approved):

PR #1716, opened 2023-12-19
Variables w code

retabak:


lessons/Early-Math/variables/langs/en-us/pages/variables-code.adoc
schanzer, line 1: Comment on the starter file itself: I'm thinking it might be better to simplify the examples, getting rid of the `==` and `true`. Can't it just be `(a + b) is (b + a)`?

lessons/Early-Math/variables/langs/en-us/pages/variables-code2.adoc
```diff
@@ -0,0 +1,71 @@
...
+|===
+
+@n Based on what you see in the Definitions Window (left side), predict what will happen when you hit "Run" by circling your choice below.
+
+@vspace{1ex}
+
+[cols="^.^1,^.^1,^.^1,^.^1", stripes="none"]
+|===
+
+| Looks shipshape, all 4 tests passed, mate!
```
schanzer, line 21: The solutions version of this question does not show an answer

lessons/Early-Math/variables/langs/en-us/pages/variables-code2.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+Who do you agree with? _Before deciding, feel free to test out Marcus's idea and Deena's idea._ @fitb{10em}{} Explain. @fitb{}{}
+
+@fitb{}{@ifsoln{Marcus's idea will work, but Deena's will not. When we test Deena's idea, examples 1, 2, and 3 pass, but example 4 fails }}
+
+@fitb{}{@ifsoln{because we cannot divide by zero.}}
+
+@fitb{}{}
+
+@star Janae dislikes Marcus's solution. She points to his code on line 17 and says, "This example is "false", so it has to be *wrong*. We can't leave it." How would you respond? @fitb{}{@ifsoln{Boolean-producing functions _should sometimes return false_! Just because an expression produces `false`}}
```
schanzer, line 67: This misconception is always a PITA for teachers to sort through. Another reason why I'm thinking we should get rid of `true` and `false` in these examples

lessons/Early-Math/variables/langs/en-us/index.adoc
```diff
@@ -167,3 +167,78 @@ If students get stuck, ask: What do you notice, at first glance? What makes the
 ** _No. If we see a representation of the Associative Property or the Commutative Property, then yes. An equation such as @math{j - y = y - j} is only sometimes true. And an equation like @math{g \times 0 = 600} is never true._
 - What does it mean for two expressions with variables to be equivalent?
 ** _When two expressions with variables are equivalent, that means that *no matter what number we substitute in,* we will still get the same result._
+
+
+== Programming Exploration: Variables
+
+=== Overview
```
schanzer, line 174: overview missing

schanzer (commented):
Overall this looks pretty good. And yay, it's all clean! ;) 

Let's discuss some of my comments in our call today. I DO think that the unplugged page with arrows has value, but I don't think you'll have time to work on it before Jan 1 so this PR may need to stay open for the next two weeks

lessons/Early-Math/variables/langs/en-us/pages/variables-code.adoc
retabak: yes! so much better!

lessons/Early-Math/variables/langs/en-us/pages/variables-code2.adoc
retabak: That's because students are making a prediction!

lessons/Early-Math/variables/langs/en-us/pages/variables-code2.adoc
retabak: makes sense.

lessons/Early-Math/variables/langs/en-us/index.adoc
retabak: will add.

retabak:
okay - something I thought you would comment on (but did not!) was the text in the lesson plan about examples.  Given that the most practice I've gotten with examples is when using the design recipe, I was unsure if I was describing them in a useful or accurate way.  Was I? I also am confused about when to use the word 'test' versus 'example'; they feel interchangeable to me, but I suspect that they are not. Can you double check that all of my dsecriptions in the lesson plan are helpful / what we want?

lessons/Early-Math/variables/langs/en-us/pages/variables-code2.adoc
schanzer: oh yeeeeah! Sorry about that :P

schanzer:
@retabak oops! I forgot to commit my changes (which include some rewriting of the examples paragraph). Just did it now, but I'll take a closer look at that paragraph and commit again if it needs more improvement

schanzer:
@retabak done!

retabak:
@schanzer - fyi, I've responded to all feedback. Take a look if you'd like! I'm keeping this PR open, hoping I'll have time to add the worksheet with the arrows. Moving onto other EM lessons now, however.

schanzer (approved):
@retabak looks great! I made a small change on worksheet 2 to improve the layout. You may want to look at the before/after just so you remember this for later worksheets, but otherwise **this is ready to Squash and Merge!**

retabak:
Noting to myself that some of the examples content is now introduced in additive inverse, so it needs to be removed here.

PR #1721, opened 2023-12-22
Additive inverse new idea

retabak:
Okay, I ran with the examples idea!  Kept the old (boring) pages but they are just optional now.

LMK what you think. If this works, I'll make another version of the same activity for multiplicative inverse.

Feel free to make any changes.

[I made more dumb merging mistakes but fixed them.... I had a brief moment of thinking this was going to merge into fall2024. Wooooops.]


schanzer (commented):
This looks great! I've pushed some minor tweaks, but feel free to push back if you don't like them. Only two issues remain:

1. There's a way to name examples _in Pyret_ so that the output in the Interactions Area is a little nicer. Take a look and decide whether or not you want to make use of it.
2. There's a minor preprocessing glitch that Dorai needs to fix. Don't let this hold you up from merging, though! Whether he fixes it now or two weeks from now, none of this will go public right away anyway.

schanzer:
@retabak our resident wizard @ds26gte has fixed the issue. Take a look at the examples-block tip, decide what you want to do, and merge when ready!

Issue #1742, opened 2024-01-10
Searchable Glossary Page


schanzer:
Just like Contracts.shtml, but uses _only the lesson URL parameter_ to display terms for a specific lesson. If none is provided, it displays all the words.

Once the feature itself has been approved, Dorai can remove all the glossary-handling code from the build script

ds26gte:
Please note that if we're avoiding creating little glossary `.html` files, we need to ensure that glossary database contains the descriptions in HTML and not AsciiDoc.

PR #1743, opened 2024-01-11
Glossary page / links

schanzer:
Implements the feature described in #1742


flannery-denny:
@schanzer 

- [ ] I think the titles for these pages need to be shorter:
- Glossary or Lesson Glossary
- Contracts or Contracts Reference Page
- [ ] I don't think Contracts belongs in Materials since it's not actually lesson-specific.  I'd suggest moving it to Supplemental Materials.  if @retabak would also like to see the glossary move there I won't object, but I'm ok with it staying since it is lesson specific.
- [ ] small nit. Glossary title is left-justified and contracts title is centered. Let's be consistent.
- [ ] For lessons without vocab words, like Computing Needs All Voices, let's leave the link out of the materials section
- [ ] For some reason some lessons didn't switch to the new format:
- Simple Datatypes
- Numbers Inside Videogames
- Making Game Images
- Functions Make life easier... 
maybe you didn't finish?

schanzer:
@flannery-denny good call on formatting. I'm not sure what you mean about "didn't finish", since none of the changes I made were in any of the lesson plans to begin with. Are you sure you've done a make clean first? If not, you're probably seeing the version of some lessons as they exist in your previous branch.

I also switched to shorter tiles as you recommended, and moved the contracts link to "Supplemental".

I'll need Dorai's help to render the glossary only for lessons with nonzero vocabulary words.

flannery-denny:
@schanzer
1 - Looks better.
2 - I don't think that the word "Pathway" means anything to teachers, so I
don't think "pathway contracts" works.
3 - Make clean resolved the lessons with old formatting - sorry I didn't
think of that yesterday.

flannery-denny:
@retabak
I disagree that we only want contracts from the current lesson in the
contracts page. This is a reference tool that students need to be able to
refer back to. Whereas, words are things we can remember and look up in
many ways, contracts are Bootstrap specific and something I still have to
look up sometimes as an author.

schanzer:
@retabak and @flannery-denny I've updated the glossary page to add the lesson title. Please `make clean` before testing, and let me know if there are any other changes you want  -- besides removing the link from lessons without glossary terms (since I need Dorai for that).

flannery-denny:
Looks great! Thanks, @schanzer. 
- [ ] The only remaining thing is that, while the page is titled **Contracts for Algebra (Pyret)** the link is titled **Pathway Contracts**. If the link title can't be course-specific, it at least shouldn't say contracts.

@retabak standing firm that we need access to the full contracts for a course and they belong in the supplemental materials section. While your notional machine of how the search bar works is glamorous, it actually only searches the words that are visible on the page (I'm not actually sure why we need it for lesson glossaries.)

schanzer:
@ds26gte this PR is ready for you! I suspect we can do quite a bit of cleanup and code removal, now that we no longer need to build temp glossary files for each lesson. All we need is to track whether a lesson declares one or more @vocab{} words, and [show the link for any lessons that do](https://github.com/bootstrapworld/curriculum/pull/1743/files#diff-144fef7c8842409835a7e1953c1ffb14ff0a2d5a5671c63194f7b35c48d6c91eR2235).

PR #1760, opened 2024-01-18
Distributive

retabak:
ready to review!

I'm getting some warnings about `Ill-named @starter-file threats` and I tried to figure out why but I'm just not sure what's going on. Sorry!


lessons/Early-Math/distributive-property/langs/en-us/index.adoc
```diff
@@ -40,121 +40,264 @@ Students develop an understanding of the @vocab{Distributive Property} by examin
...
+@Q{Does `scale` distribute over `beside`? How do you know?}
+@A{Yes, it does. We know because both Circles of Evaluation produced identical images.}
+}
+
+@lesson-instruction{
+- Complete @printable-exercise{distribution-and-code.adoc}.
+}
+
+@QandA{
+@Q{What did you discover about the distributive property and code?}
```
schanzer, line 262: Curious what you expect students to reply with, here. Might be helpful to include a sample `@A{...}`

lessons/Early-Math/distributive-property/langs/en-us/index.adoc
```diff
@@ -40,121 +40,264 @@ Students develop an understanding of the @vocab{Distributive Property} by examin
...
+
+=== Investigate
+
+Let's look at whether we can distribute two additional functions: `flip-vertical` and `flip-horizontal`.
+
+An interesting feature of these functions is that they require just *one* input: an Image. In a mathematics context, we have applied the Distributive Property to functions that require two inputs... but never one. Let's give it a try now!
+
+@slidebreak
+
+A function `triple` takes a single value and multiplies it by three. Study the pair of Circles of Evaluation below to see if we can distribute the function `triple` over addition.
```
schanzer, line 275: I like this a lot, but I think you can make it even stronger by introducing `triple` _before you get to programming_.

1. Tell them there's a made-up operation called triple, and you want to know if it can be distributed over addition. 
2. Then remind them of it here in the programming section

If we don't show that this is based _sound mathematics_, we'll have to deal with teachers who claim this is all just "programming stuff, not math"

lessons/Early-Math/distributive-property/langs/en-us/index.adoc
```diff
@@ -40,121 +40,264 @@ Students develop an understanding of the @vocab{Distributive Property} by examin
...
+
+Extending concepts explored earlier in the lesson, students consider whether various functions that we use when coding are distributive.
+
+=== Launch
+
+We've learned a lot about how the distributive property works. We know that multiplication can distribute over addition or subtraction, allowing us to complete otherwise complex computations with efficiency and ease.
+
+@slidebreak
+
+But what about our programming functions? For instance, does `scale` distribute over `beside`?!
```
schanzer, line 234: There is no such thing as a "programming function" in Pyret or WeScheme. They are simply **functions**. I suggest rewriting this as "But what about functions that deal with images?"

The more you make "programming" sound different from math, the more ammunition you give math teachers to resist. I can just hear someone saying "maybe that's cool for programming functions, but this is a _math class_. Why should I make time for something that isn't math?"

lessons/Early-Math/distributive-property/langs/en-us/pages/distribution-and-code-2.adoc
```diff
@@ -0,0 +1,98 @@
...
+  table {grid-template-rows: 3fr 1fr !important;}
+  div.circleevalsexp .value,
+  div.circleevalsexp .studentBlockAnswerFilled { min-width:unset; }
+  .sect1 > h2:first-child { padding: 0 5px !important; }
+  .sect1 > h2 { font-size: 11pt !important; }
+  .autonum { padding: 0 !important; }
+</style>
+++++
+
+Open the @starter-file{comm-and-assoc}, which you will use to investigate three functions:
```
schanzer, line 14: I see four functions :)

lessons/Early-Math/distributive-property/langs/en-us/pages/distribution-and-code.adoc
```diff
@@ -0,0 +1,106 @@
...
+  table {grid-template-rows: 3fr 1fr !important;}
+  div.circleevalsexp .value,
+  div.circleevalsexp .studentBlockAnswerFilled { min-width:unset; }
+  .sect1 > h2:first-child { padding: 0 5px !important; }
+  .sect1 > h2 { font-size: 11pt !important; }
+  .autonum { padding: 0 !important; }
+</style>
+++++
+
+Open the @starter-file{comm-and-assoc}, which you will use to investigate three functions:
```
schanzer, line 14: Four functions

schanzer (changes requested):
I left some small comments that are easy to fix, but I'm left with a larger meta-comment.

The "Does X distribute over Y?" worksheets are _great!_ Is there any way to duplicate them in the math section? Like once they've seen multiplication distribute over addition, maybe they could do an identical worksheet that's strictly math. "Does triple distribute over addition?" would be cool, and you could make up 2-3 other functions as well. That would make the programming stuff WAY less "new" and WAY less intimidating.

ds26gte:
As I said on Slack: the variable Projects is defined twice in starter-files.js. The second definition overwrites the first. The first one contains the entry for "threats", but since it's overwritten, we get an ill-defined warning for "threats".

schanzer:
Thanks for the reminder, @ds26gte ! @retabak , I have no idea which one of these is the right one, but I think you have a better sense for which one needs to be deleted.

lessons/Early-Math/distributive-property/langs/en-us/index.adoc
retabak: you're so right.

lessons/Early-Math/distributive-property/langs/en-us/index.adoc
retabak: heard. This makes good sense.

lessons/Early-Math/distributive-property/langs/en-us/pages/distribution-and-code-2.adoc
retabak: woops!

retabak:
@ds26gte thanks for discovering this problem! I am seeing that I have misinterpreted the merge conflicts that appear in the starterfiles.js and handled them poorly. Will try my very best to avoid this in the future (or at least look for it as the cause of a mysterious warning!).

PR #1762, opened 2024-01-19
Identity prop

retabak:
I reworked this completely. I'm not sure if it works. I stole/modified some role play from our functions lesson.... it seemed promising but I am not sure if it works / is helpful, either. Some of my language can definitely use some polishing too. The lesson is dramatically different, but the worksheet is actually not very changed at all. LMK what you think. Have not updated slides yet because I'm not totally confident in what I have here.


lessons/Early-Math/identity-property/langs/en-us/index.adoc
```diff
@@ -167,33 +167,104 @@ Students explore two functions - `scale` and `rotate` - each of which consumes a
...
+
+| @show{(coe '(* m 1))}
+| @show{(coe '(/ m 1))}
+| @show{(coe '(+ m 0))}
+| @show{(coe '(- m 0))}
+
+|===
+}
+
+Each of these four Circles evaluates to @math{m} which means *the Identity Property requires an operation that does… nothing!*
```
schanzer, line 178: I LOVE this table and the summary statement! If it were the late 90s I would demand that you make a Seinfeld reference "It's arithmetic about _nothing!_"

I feel like this belongs in the Synthesize section of "The Identity Property and Variables." Besides, didn't we already agree that one-input identity functions that work with numbers belong firmly in the _non-programming_ section?

lessons/Early-Math/identity-property/langs/en-us/index.adoc
```diff
@@ -167,33 +167,104 @@ Students explore two functions - `scale` and `rotate` - each of which consumes a
...
 }
 
 
+No matter what value we use for @math{m}, we will get that value back.
+
+@slidebreak
+
+
+@lesson-roleplay{
+Select four students - one to _act out_ each of the four functions above (`multiply-by-one`, `divide-by-one`, `add-zero`, and `subtract-zero`). Make it clear to the class what each function's name is. Emphasize that each function expects a Number, and will produce Number.
```
schanzer, line 205: I love this idea, but I worry that it will get boring too quickly. Could you add another twist, which is that the class gets to brainstorm _other_ "actors" who follow the identity property? This would be a nice callback to some of your "tricky" examples in the prior section, like "a * (8 - 8)".

lessons/Early-Math/identity-property/langs/en-us/index.adoc
```diff
@@ -167,33 +167,104 @@ Students explore two functions - `scale` and `rotate` - each of which consumes a
...
+@teacher{
+Solicit student responses and record them on the board. If students do not volunteer answers, consider sharing one or two of the following possible responses: flip it vertically twice; flip it horizontally twice; rotate it 360 degrees clockwise; rotate it 360 counter-clockwise; scale it by 1; slide it some distance and then return it to its original position.
+
+Students will be working in @proglang soon - but we recommend keeping things unplugged for now! You might even print up an image of the dog - or any other image you choose - to display on the board at the front of the room. Invite students to the board to demonstrate their ideas by manipulating it.}
+
+@slidebreak
+
+
+@lesson-instruction{
+- Open @starter-file{identity-prop} and click “Run.”
```
schanzer, line 244: The starter files defines `hello`, which is a text image. `dog` is unbound. Maybe you need to hit publish again?

Overall I love the way you lay this out, and I'm really happy with your worksheet

schanzer (changes requested):
This is a MAJOR improvement. I'm really liking this lesson!

I made a bunch of small formatting changes in the wbpages, designed to give kids more room to write their open-ended explanations. I think you're over-using the @QandA, @teacher, and @lesson-instruction blocks at the beginning, which makes the lesson seem very busy without a ton of payoff. I tried to consolidate them a bit, but let me know if you think I've overdone it.

Finally - the starter file doesn't contain `dog`, so I can't test the worksheet. That's the only real show-stopper here. Once you've resolved that, feel free to merge if you're ok with my changes.

lessons/Early-Math/identity-property/langs/en-us/index.adoc
retabak: we previously talked about one-input functions (i.e., triple) with respect to the distributive property, not the identity property.... but it seems like this is a recurring theme! I like this feedback.

Issue #1759, opened 2024-01-18
Add a Lesson and Standards overview


flannery-denny:
We'd ideally like one list of standards by lesson and another list of lessons by standard as seen below from open up
![image](https://github.com/bootstrapworld/curriculum/assets/87144723/e240ae32-4c9a-4910-acc2-daf32d057446)

schanzer:
Closing, now that there's PR for it (#1768)

PR #1768, opened 2024-01-22
New alignments page

schanzer:
This PR addresses #1759


flannery-denny:
Wow. Feels like magic! Thanks for the quick turnaround @schanzer 

At a base level, I think this is super useful and impressive and we could add it to Teacher Resources as is.

From an aesthetic point of view I think a few small tweaks would make it much nicer.

When we start with a lesson, I love that it's very neat and organized. And I wonder could we reduce how much bolded purple there is? Maybe the Alignment dictionary titles could be italicized or blue instead? or the lesson titles could be underlined instead?
Screen Shot 2024-01-23 at 7 12 18 AM

When we start with a standards dictionary, I am impressed by how many lessons we have addressing various standards! And I wonder could we delineate a little more clearly between the standard and the alignments. Perhaps just starting [See: CC56871, etc.] on a fresh line would be enough? Or perhaps we want 2 columns like open up usese?

Screen Shot 2024-01-23 at 7 12 33 AM

schanzer:
@flannery-denny I've made the changes you asked for!

flannery-denny:
@schanzer I think this looks way better! Remaining todos before I sign off:

1. The search bar isn't working for me.
2. Now that the lessons are on a new line, I think we can get rid of `[See: ]` and just list the lessons. Thanks!
Screen Shot 2024-01-23 at 11 23 30 AM

schanzer:
@flannery-denny and @retabak I've removed the italics to make the rest plain, gotten rid of `[See:`, and renamed the Math practice standards.

retabak:
all of these small fixes look good and the search bar seems fully functional.... but it is oddly positioned when viewing by alignment:

image

I much prefer the formatting here:

image

PR #1931, opened 2024-03-01
Variables w code

retabak:
PLease review new worksheet, which needs to be integrated into the lesson... POSSIBLY in the coding section, but possibly elsewhere?


schanzer:
@retabak this is EXACTLY what I had in mind. Nice work! I spent some time figuring out a way to make solutions work -- it's a hack, but it should be easy to follow. :)

I don't think this belongs in the programming section, though - there's no code involved!

retabak:
@schanzer - I believe this is ready to merge. Take a final look when you have a moment and LMK.

schanzer:
@retabak merge away!

PR #1945, opened 2024-03-06
Notes and slides (fixes #1725)

### LINKED ISSUE #1725: Replace @lesson-notes with @lesson-info
Spec! (3/5/24):
--------------------------------------------------
@flannery-denny and I (Rachel) met and we have a spec for @lesson-info!

Here is what we came up with (screenshot from a google doc):
image

- [ ] [Chosen Dataset ] links to dataset table

Updating description (1/19/24):
--------------------------------------------------
This should _still_ happen on the [notes-and-slides branch](https://github.com/bootstrapworld/curriculum/tree/notes-and-slides). :)
- [x] Rename directive to @lesson-info
- [x] Have it simply produce an empty div or span with the ID `lesson-info-table`. If this is onerous, we can make do with it being the className
- [x] @ds26gte will add new fields to the lesson-graph: `slides` and `notes`. 
- [x] @schanzer will write some JS to create this table based on the dependency graph

Updating description (1/6/24):
--------------------------------------------------
This should happen on the [notes-and-slides branch](https://github.com/bootstrapworld/curriculum/tree/notes-and-slides)
- [x] Rename directive to @lesson-notes-and-slides
- [x] Instead of displaying a list of notes pages, display a 3-col table (lesson name, slide link, notes if they exist)
---------------------------------------------------

Would it be possible to add a section to Teacher-Only Resources that would house all of the lesson slides in one place?  It would look like the All Lesson Notes dropdown.  I had a teacher ask for it.  I could not find it anywhere, so I created [this document](https://docs.google.com/document/d/1ot_w5oypZeoU4_VhT2jmRzp8NQqIebxNdXbz7Z0-Rjw/edit?usp=sharing) housing all of the lesson slides.  She has independent students working through the curriculum and wants them to use the slides as a self-teaching tool and to get links to all workbook pages, so having them all in one place was needed.  I created and shared this doc with her, but perhaps it would be good to have it in our resources.

schanzer:
This adds a lesson info table, as spec'd in #1725


retabak:
yay! Eager to check it out.

retabak:
@schanzer - can you specify where to look to view this? I've checked out the branch `notes-and-slides`, pulled, then I did a clean make... but when I go to any pathway's Teacher Resource Page (where I thought this table would appear?), I just see old stuff.

When I looked at the algebra-pyret teacher resource asciidoc, I discovered that `@all-lesson-notes` was still there! So I replaced that with `@lesson-info` ... and nothing appeared.

Am I missing something totally obvious? How can I see what I wanna see?

schanzer:
@retabak arg! I forgot to check in a file. Fixed. Look in the Algebra-Pyret pathway, TR page

retabak:
@schanzer - Did a fresh pull and clean make. Still not seeing anything.

schanzer:
@retabak Are you sure you're looking at the right URL? Go to the algebra-pyret main page, and click "Teacher Resources". Then expand the "notes and slide decks" section.

retabak:
Oh. That worked. I was opening the teacher resource page from the algebra-pyret resources folder in distribution. I don't entirely understand why that didn't work? I'll take a look at this.

schanzer:
@retabak opening a lesson by itself shows you a reduced version of the lesson, so that doesn't work either! That's how it's worked since fall 2022.

retabak:
Just realized I gave my last feedback in #1725 so am copying it here for the record.

-----------------

2nd feedback round:

-[ ] each row should be the same height, if possible. This might require making the first column narrower to give the third column more space (since the starter files occasionally use 3 lines of text... I'm worried we won't be able to fit all of the flag starter files, regardless... boo! Maybe we need to rename some?)
-[ ] Can the text have top vertical alignment rather than center vertical alignment? 

I'm not sure how I feel about starter files in brackets, either. @flannery-denny , wanna weigh in?

-------------------

1st feedback round:

Okay, before I give some very nitpicky feedback, it might help to know that @flannery-denny and I both like how [this table](https://www.bootstrapworld.org/materials/fall2023/en-us/lessons/choosing-your-dataset/pages/datasets-and-starter-files.html) looks and feels - so if there's any formatting you can grab from there, you should!

- [x] make row height shorter, while still keeping all rows the same height
- [x] put "Slides" and "Notes" inside brackets (e.g. [Slides], [Notes])
- [x] put starter file titles inside brackets
- [x] starter file names separated by 2 spaces and no comma
- [x] third column text should be right-aligned
- [x] remove the ourside border from the entire table

This is super annoying, I'm sure. Sorry!

schanzer:
@retabak and @flannery-denny I've responded to the 2nd round as best I can. There is no way to make all the rows the same height without making them all as tall as the tallest row...which is a _ton_ of wasted space, and even then it's brittle because someday we might have more starter files in a lesson...and the extra line will be cut off until we find it and raise the fixed-row-height. I strongly advice against this.

As a compromise, I've made all the rows 2 lines tall. This makes 90% of the rows the same height, with minimal wasted space.

flannery-denny:
@schanzer The tables aren't building at all for me.

retabak:
Assuming you saw this comment from @schanzer?

*Are you sure you're looking at the right URL? Go to the algebra-pyret main
page, and click "Teacher Resources". Then expand the "notes and slide
decks" section.*

flannery-denny:
Thanks @retabak. I can see the table now... not sure why `distribution/algebra-pyret/resources/index.shtml` wouldn't get me to the same place.

1. Sorry I was so far down an authoring rabbit hole that I missed the pull review before changes were made.
2. I think the brackets around the links are overwhelming.
3. I went back to the table in Choosing Your Dataset to see why we liked it / was the formatting slightly different? Turns out it actually uses a combination of `[` `|` `]` and more ample horizontal space.
Screen Shot 2024-03-08 at 12 45 48 PM

 I think it would actually be tough to implement that formatting here since there are line breaks to work with.
--> I would suggest removing the `[ ]` and adding ` | ` between starter files instead (note the space before and after)
4. Noticing that the Desmos file names are so long that they break lines and appear to just say "Desmos"
Screen Shot 2024-03-08 at 12 53 18 PM
--> Could we make sure names to break across lines?
--> Since there should only one Desmos activity per lesson now, should we shorten all of these titles to say Desmos Activity?

schanzer:
@retabak no need. That's what this thread is for. Done!

retabak:
Dang, that was quick. It hadn't occurred to me that we could do find + replace :)

flannery-denny:
@schanzer spoke to @retabak and we've agreed to add this issue to the curriculum team agenda and end the back and forth.

To help us remember our ideas... 
- we believe it's important that no link title appears across two lines
- we think Desmos activity links should all read "Desmos Activity"
- we wonder would shrinking the font help the table formatting?
- what other styling options are available to us besides everything have `[ ]` around it? Let's at least explore an option with `|` as is used in the dataset table.

Also, want to recognize that while we haven't totally nailed this yet, it feels close and it's clear that the end result will be a huge improvement for teachers looking for an overview of available materials.

PR #1971, opened 2024-03-13
Simpler table indexing

schanzer:


schanzer:
@retabak  and @flannery-denny it looks like we have a winner! Confirming this correctly processes the functions-can-be-linear lesson. If you have 15m to block off checking 3-4 lessons and comparing them to the same slides generated from fall2024, I think we've got a much simpler, more stable mechanism for getting the right tables on the right slides.

retabak:
Sorry I didn't get a chance to look! I trust Flannery! You should merge.

PR #2076, opened 2024-06-10
updates to exploring quadratics and some more linear formatting

flannery-denny:
There isn't anything that doesn't work about this lesson, but we really need to find another quadratic dataset so that they can apply what they've learned.

Any chance there's fuel efficiency data available for other kinds of vehicles that's also quadratic?


schanzer (changes requested):
@flannery-denny overall looks terrific. Do a bulk find-replace for maximum and minimum to switch over to maxima and minima, and we're good to go!

flannery-denny:
@schanzer fyi that involved changing a glossary definition.

PR #2083, opened 2024-06-12
Logarithmic Updates

flannery-denny:


flannery-denny:
To Do:
- change `pc-gdp` to be "per capita gdp in thousands of dollars" in the source file.
- decide whether this whole slider activity (`exploring-logarithmic-models/pages/graphing-models.html`) can happen in base 2 or there's a good reason @schanzer had the last exploration switch to base 10 (which means the curve can't be compared to the underlying curve)
- find out whether @ds26gte can easily make `@math` render in page titles

schanzer:
@flannery-denny 

1. Your intuition was right! the 4th desmos slide (which accompanies _graphing-models_) was saved with the coefficients set to the wrong place. This has been fixed.
2. After playing with this, there is _definitely_ a reason: the interplay between _a_ and _b_ is subtle and important, and seeing it requires some quick arithmetic. Students are far more likely to be comfortable working with powers of 10 than powers of 2 (or anything else). I suppose we _could_ make g(x) use base 10 as well...I'm not opposed to that. Making x go to a million was also unnecessary, so I've knocked this down to 1,000 which is far more manageable.
3. Up to the wizard. 😉 

I also see that the desmos activity had slide 5 set to use _natural log_, which is flat-out wrong. I'm guessing that I did some playing around during the last PD, thinking I was in preview mode...and then accidentally wound up changing the activity itself. This has been fixed.

flannery-denny:
@schanzer realizing that in order to change `pc-gdp` to `pc-gdp` in thousands for pyret, I not only have to change the google sheet, but also have to change the desmos files.
These are the new files:
- [google sheet](https://docs.google.com/spreadsheets/d/1B6Pg-kcPoxkVD5MOGzS_ABrLS6hfLSfcHTYY0cHm2kU/edit?usp=sharing ) - question: Is it worth keeping the original pc-gdp and adding a new column for pc-gdp-thousands, which we'll use in the model? or should I delete the OG pc-gdp column
- [desmos](https://teacher.desmos.com/activitybuilder/custom/666b05753ceb10c0827de525) - I've updated the tables, the x-axis, and labelled all of the axes.
-- I  haven't updated the second table on slide 6 yet because I haven't actually read the accompanying materials to understand what this slide is supposed to be doing yet. 
-- Similarly, I'll have to think my way pretty deep into this lesson in order to intuit where the sliders are intended to be on these slides.

Refocusing my energies on the exponential unit for now. Will let you know if I come back to working on these today. Otherwise, anything you do to polish the sliders on this activity so that they will make as much sense as possible and I might be able to update workbook pages on a first pass is welcome.

flannery-denny:
@schanzer You haven't weighed in yet on whether it's worth keeping the original `pc-gdp` and adding a new column for `pc-gdp-thousands`, which we'll use in the model,  so I am making an executive decision to update the file to include both.

flannery-denny:
actually @schanzer I thought that I was going to get to the work related to updating `pc-gpd-thousands` today, but I need to finish dealing with the lesson at a more zoomed out level, so there's time for you to weigh in.

schanzer:
@flannery-denny apologies - I've gotten pretty sick over the last few days and it's been harder to keep up.

I think keeping both is a great idea!

flannery-denny:
@schanzer logarithmic is getting closer!

Remaining todos.

- [ ] Add solution for *Why did transforming the *x-axis* make our data look linear?* If wording for this comes easily I'm all ears
- [ ] Add some directions into the *fitting by changing scale* section about toggling back and forth between linear and logarithmic axes 
- [ ] Need make sure all solutions pages are updated with values and wording that correspond to `pc-gdp-in-thousands` switch (the function names has already been swapped out, but there's more to it than that!)
- [ ] Tackle the *Fitting By Transforming Data* section 
- [ ] Update Pyret screenshots in *Fitting By Transforming Data* section 
- [ ] There are a ton of pages in this lesson that don't fit on a page. Haven't looked yet to see whether this means they need to be split or I can clean it up with CSS

flannery-denny:
A few places that need us to talk about log base 10

- In Fitting by Changing the Scale
Screenshot 2024-06-19 at 8 58 10 AM

- On Changing the Scale we have these two definitions side by side (one with 10, one without)
Screenshot 2024-06-19 at 8 58 37 AM

flannery-denny:
@schanzer delegating this to you. Thanks!

- [ ] please update the second table on slide 6 of the health-v-wealth-transformed Desmos file (currently owned by teachersuccess! the other version doesn't have `gdp-pc-in-thousands` data) 
- [ ] While you're in there also please change `h(x)` to `f(x)` on slide 4

flannery-denny:
@schanzer More thoughts to discuss in our meeting re: this lesson

- [ ] Noting that the S-values don't get that much better in the logarithmic models lesson, yet we write " showing a vastly better correlation than before." The lesson plan also states "showing a much better fit than our 2-point-derived estimates." We did 2-point derived estimates in the linear lesson, but that's not part of this lesson...
- [ ] Logarithmic Interpretation stem.
- [ ] In desmos we switch from "linear" scale to "logarithmic" scale. In the lesson plan we often say, "By transforming the x-axis to grow exponentially," Do we want to talk about the relationships between these two words or just be more consistent in using logarithmic?
- [ ] From the fitting my transforming data section of the lesson plan
- Were those values very similar or very different? TO WHAT
- How close were the values from our best-guess logarithmic model? WHICH VALUES
- [ ] I see the reason to compare the transformed table in desmos but is there a good reason to fit the linear model in Desmos before fitting it in pyret?
- [ ] Can we turn this into QandA instead of narrative. 

> Transforming the points instead of the axis has the same visual effect: the dots appear to fall in a straight line. But now we can plot them on a linear-scale axis, and use linear regression to find the best-possible model!
- [ ] I have had to change the vertical shift on all of your solution pages from ~24 to ~59. It's a big change. And I just want to triple check that nothing has gone egregiously wrong in the data somewhere
- [ ] Is there a good reason to write the lr-plot definition in pyret notation as a linear function? I'd prefer to just use y= for this function since that's what it actually says on the lr-plot and we're not going to add it to pyret until we use the coefficients in the `logarithm3(x)` function definition in pyret.
Screenshot 2024-06-20 at 11 22 45 AM


- [ ] Let's discuss part 2 of logarithmic models

schanzer:
Things we discussed in our meeting:


- [ ] Add question to logarithmic models page(s) that asks students to compare the _improvement in S_ relative to other models
- [ ] We need to add an interpretation stem for logs
- [ ] Need to use the word "exponential" when describing the x-axis tick-marks, either in wbpage or lesson plan
- [ ] Revise "Transforming the Data" page and accompanying section of lesson plan
- [ ] in "logarithmic models", students shoulnd't have to run `fit-model` on the linear function....`lr-plot` already gave them the S!

flannery-denny:
@schanzer I’m still struggling with your inconsistent use of exponential / logarithmic as what we changing the scale from linear to. For an example see the launch of fitting my transforming data.

PR #2090, opened 2024-06-20
periodic updates

flannery-denny:
TODAY:

- [x] FD adds a periodic vocab page, using the graph image from the lesson plan
- [x] FD adds desmos instruction prose (desmos file use)
- [x] FD will changing modeling-ferris-wheel wbpage to specify the two functions kids need to write
- [x] ES ...Emmanuel TODO...Need a wbpage for converting between "descriptions", "clock time", "degrees", and "radians"?

DONE
- [x] ES - Graphing periodic models - change desmos scale so periods and maxima/minima are on major gridlines 
- [x] ES will add `modern-table` and rename `recent` to `recent-table`, then revise the lesson plan to start with modern and then zoom out.
- [x] ES - when showing the two periodic models in the lesson plan, call them `periodic1` and `periodic2`
- [x] ES - add "how is .375 of the year in _May?"_ section
- [x] ES we need an interpretation stem for when they merge
- [x] ES We need an interpretation stem for periodic models somewhere
- [x] ES - address the fact that in pyret `PI / 2` needs spaces around the fraction bar
- [x] ES splits the x/y coordinates on unit-clock
- [x] ES will turn make-a-wheel into a 2-section page
- [x] ES - shift the y-axis left on these grids Screenshot 2024-06-20 at 2 18 10 PM
- [x] add 2graphs of non sinusoidal periodic functions to the a note on vocabulary box
- [x] Students will need more context for the C02 table date since it's by day but the only other data in the table is the month - making the decimals not intuitive. for example, I would expect the decimal for may to be 5/12 which is .416667 but somehow it's .375
- [x] for synthesize section of come up with at least 1 example of a real world periodic functions that they can do the second question with before they come up with their own examples
- [x] rewrite this
Screenshot 2024-06-20 at 1 55 06 PM

- [x] is it intentional to avoid the words maxima and minima in the graph vocabulary?
- [x] also seeing us be inconsistent about `height` v `altitude`
- [x] This should not be the very first mention of sine and cosine
Screenshot 2024-06-20 at 2 15 28 PM

- [x] I'm not buying that any algebra2 teacher would do that from radians to degrees kinesthestic activity. How about a simple diagram explaining why half a circle is equivalent to pi radians and the link to the video? 
Screenshot 2024-06-20 at 2 05 21 PM

- [x] Fix axis labels on graph to match table on Exploring Periodic Data
- [x] Fix axis labels on graph to match table in lesson plan
- [x] starter file does not match Exploring Periodic Data data
- [x] A and B are used inconsistently... sometimes as positives... sometimes as real numbers. let's discuss image
- [x] Do alg2 teachers expect students to rationalize the denominator?
Screenshot 2024-06-20 at 1 40 49 PM


schanzer:
From our discussion:

- [ ] In Periodic models, replace height with altitude 
- [ ] in lesson plan and wbpages, make sure axes are for time and altitude
- [ ] Remove “got ot pyret” from line graphs in exploring ferris wheel
- [ ] Should we rationalize the denominator? YES
- [ ] Flannery will merge (x,y) and lengths of A, B in the workbook page and lesson plan
- [ ] Emmanuel will add discussion of cos and sin being functions after “Reaonsing about Unit Circles” wbpage
- [ ] Take out string and paper-plate activity

flannery-denny:
- [ ] Is there a good reason to add about lowest and highest before making the line graph? Seems like it should go in the other order.
- [ ] let's explicitly ask what pattern they see and what ride they think the teacher was on before connecting the dots
- [ ] remind me to tell ES how desmos links are supposed to be used in our materials
- [ ] explicitly call out the pythagorean theorem in QandA before setting students loose to calculate x and y values of unit circle
- [ ] I vote we make the table for unit circles have an x-coordinate column and y-coordinate column
- [ ] we still have to talk about sine and cosine before introducing the pyret functions - lesson points aren't supposed to introduce brand new ideas
Screenshot 2024-06-21 at 8 47 05 AM

- [ ] we definitely need to add diagrams when describing slicing the circle in other ways
- [ ] In fact we need a workbook page that engages students with radians before that graph that uses radians
- [ ] I would suggest offering up this image on a workbook page and asking students to take a minute to try to define the words used on the diagram
periodic graph

- [ ] I don't think `f(x)` is serving us in introducing the periodic models.
- [ ] If we're going to say that the models are the same just shifted and equally good, I think we need to explicitly draw that out of students earlier in the lesson
- [ ]

flannery-denny:
@schanzer In case it's better to see this here:
I got overwhelmed by the function renaming because the lesson plan doesn’t seem to be linking to the Teacher Success version of the Pyret files and I couldn’t quite wrap my brain around how all of the functions in the Pyret files / Desmos files / other lesson materials intersect

I propose:
- periodic1 and periodic2 for ferris wheel
- periodicA, etc. for Co2
- Unless desmos slides are supposed to be referring back to each other, new function names on every slide.
Perhaps you could attempt to sort this out?

(for example, on modeling the ferris wheel, I’m not actually clear whether the f and g they are changing in desmos were supposed to be the same f and g from pyret - I think not, but can’t quite sort it.)

PR #2180, opened 2024-09-05
Assessment branch - all 4 lessons updated with links to Desmos FAs

retabak:


lessons/Data-Science/box-plots/langs/en-us/index.adoc
```diff
@@ -34,7 +34,10 @@
 |
 @preparation{
 - There is an optional kinesthetic activity in this lesson that requires a ball of play-dough for each group of 3.
-- There is an optional interactive Desmos activity in the lesson. If you would like to use it, decide how you will share it with students and, if you are using our Google Slides, add the appropriate link to the slide deck. _If you're a first-time Desmos user, fear not!_ @dist-link{pages/desmos-instructions.adoc, _Here's what you need to do._}
```
flannery-denny, line 37: @retabak 

1. What happened to the Desmos activity that the preparation section originally warned teachers about?
2. The preparation section should definitely acknowledge the cumulative assessment.
3. Looks to me like you've added 4 new desmos files to this lesson, but only linked two of them from the preparation section? If we're already struggling to maintain the links, wondering if we actually need an 
| Assessment section with
| [.assessment links]
and an @assessment{} directive that refers to a library that isn't part of `starter-files.js` (I actually had previously misunderstood that y'all had already made a separate library)
Thoughts @schanzer?

retabak: 1. I meant to change this text to look like the histograms lesson preparation text, e.g., "There is an interactive Desmos activity in the lesson, as well as Desmos formative assessments" - but apparently never finished. Thanks for catching this. (That desmos activity is still, of course, in the lesson.)
2. Could I just say, "There are Desmos assessments at the end of each lesson section"? If I mention the cumulative assessment, I would feel obliged to say that it just combines the formative assessments... but then we'd be adding a lot of text. I'd feel more comfortable adding this if it were in its own section, which leads me to item 3...
3. This is a great idea, and I think we should do it!

schanzer: @flannery-denny @retabak let's open a separate issue to track assessments in a dictionary, and fix this by hand for now. No point in holding up the PR.

Issue #2263, opened 2024-10-31
add section on image bar and pie chart to lookups


flannery-denny:

schanzer:
Closing this, as it's now subsumed by #2303

PR #2303, opened 2024-12-04
Table Functions Pathway

schanzer:
In tackling #2263 , I came to the realization that the larger issue is a lack of continuity that ties together Defining Functions (Core), Lookups (DS), Defining Table Functions (DS), and Filtering and Building (DS). This lack of harmony made functions like `image-bar-chart` and `image-pie-chart` stick out like sore thumbs, because they suddenly required an integrated understanding of all these disjointed parts.

The original spirit of #2263 was to build a bridge from `lookups` these functions, by introducing concrete lookup-using expressions at the end of the lesson that would then feed into the `image-*-chart` functions. Unfortunately, this proved to be a bridge too far, because only after reifying those concrete expressions as _functions_ could they be used with `image-bar-chart`. 

What we want is a clear through-line:

1. use contracts to make simple expressions ("make a solid green triangle!")
2. use table contracts to introduce row lookups ("get the `age` of the **7th** row!")
3. use lookups to make expressions from rows ("make a solid green triangle whose size is the `age` of the **7th** row!" or "check if the `pounds` of the 7th row is greater than 50!")
4. use expressions from rows to make row-consuming functions ("write a function that consumes **any** row and makes a solid green triangle using `age`!" or "write a function that consumes **any** row and checks if `pounds` is greater than 50!")
5. use row-consuming functions for filtering and building ("filter the table to show only animals over 50 pounds! build a column showing the age of each animal as a solid green triangle!")
6. use row-consuming functions for making advanced displays ("make a scatterplot of `pounds` vs. `weeks`, where `age` is shown via a solid green triangle")

This PR implements the following:

- [x] Bring the Core `contracts` lesson into DS pathways
- [x] Change DS `contracts-tables-visualizations` to remove content that's now covered by `contracts`, and replace it with material about `row-n`. 
- [x] While we're at it, make better use of `table-contracts` workbook pages
- [x] Change `lookups` to not spend as much time on `row-n`, since this is review instead of a thorough introduction
- [x] Make better use of `defining-rows` workbook page, which now sets up a new `lookup-expressions` page that gets students writing concrete lookup-using expressions on those defined rows.
- [x] Base `defining-table-functions` on the concrete expressions from the preceding `lookups` lesson, in which those expressions become reified as abstract functions
- [x] Revise `filtering-and-building` to use those same functions
- [x] Revise `advanced-displays` to use those same functions
- [x] Splits `contracts-tables-visualizations` into two lessons
- [x] Confirm Algebra 2 works without needing multiple contracts lessons


lessons/Core/contracts/langs/en-us/index.adoc
```diff
@@ -342,7 +342,8 @@ It can be empowering for students to develop expertise on a topic and get to sha
 
 @lesson-instruction{
 - Let's do some more experimenting with some of the more complex image functions from the list we just saw! This time around we'll start with their contracts.
-- Complete @printable-exercise{triangle-contracts.adoc}, @printable-exercise{using-contracts.adoc} and @printable-exercise{radial-star.adoc} with your partner.
+- Complete @printable-exercise{triangle-contracts.adoc}, @printable-exercise{using-contracts.adoc} with your partner.
+- @opt{If time allows, complete @opt-printable-exercise{radial-star.adoc} with your partner.}
 }
```
flannery-denny, line 348: Confused why you want radial start to become less discoverable. Objections to adding it to the end of the opt directions that follow?

lessons/Data-Science/contracts-tables-visualizations/langs/en-us/index.adoc
```diff
@@ -45,11 +45,11 @@
 
 |===
 
-== Functions for Working with Tables @duration{20 minutes}
+== Functions for Tables @duration{20 minutes}
 
 === Overview
 
-Students will be introduced to functions for working with tables, including: @show{(code 'sort)},  @show{(code 'count)} and @show{(code 'first-n-rows)}. In the process they will review simple @vocab{data types} and be introduced to @vocab{contracts} for describing the @vocab{Domain} and @vocab{Range} of any function.
+Building on their understanding of @lesson-link{contracts, Contracts}, students will be introduced to functions for working with _tables_, including: @show{(code 'sort)},  @show{(code 'count)} and @show{(code 'first-n-rows)}.
```
flannery-denny, line 52: add row-n

lessons/Data-Science/contracts-tables-visualizations/langs/en-us/pages/functions-for-tables.adoc
```diff
@@ -7,70 +7,67 @@
...
+@n Think of another question you might be able to answer with the `count` function.
 
-@fitbruby{10em}{@ifsoln{sort}}{function-name} ( @fitbruby{20em}{@ifsoln{animals-table}}{table-name {two-colons} Table},  @fitbruby{15em}{@ifsoln{"legs"}}{column-name {two-colons} String}, @fitbruby{10em}{@ifsoln{true}}{order {two-colons} Boolean})
+@fitb{}{@ifsoln{Possible response: How many animals are cats?}}
 
-@n Which animal(s) had the most legs? @fitb{}{@ifsoln{Snuggles the tarantula}}
+@n Fill in the blanks with the code you'd write.
 
-@n Think of another question you might answer quickly by sorting the table.
+@fitbruby{10em}{@ifsoln{count}}{function-name} (@fitbruby{20em}{@ifsoln{animals-table}}{table-name {two-colons} Table}, @fitbruby{15em}{@ifsoln{Answers will vary!}}{column-name {two-colons} String})
```
flannery-denny: @schanzer The rendering of this is inconsistent. Sometimes it works. But most of the time I see "two-colons" and the datatype appears on the fitb.

lessons/Data-Science/contracts-tables-visualizations/langs/en-us/index.adoc
flannery-denny, line 1: @schanzer I would suggest that we make this page from the composing circles required rather than optional:

@opt{If your students would benefit from seeing a few more examples before drawing their own Circles of Evaluation, have them complete @opt-printable-exercise{matching-coe-to-descriptions.adoc}.}

lessons/Data-Science/lookups/langs/en-us/index.adoc
flannery-denny, line 1: @schanzer this lesson starts off with the same exact stuff that's used in the tables, rows and visualizations lesson. (lines 37-50) ... but in the other lesson the discussion is more advanced in that it discusses`row-n`. That doesn't make sense to me. I'm also seeing the entire first section of this lesson repeat within itself. 

Now that we've added row-n to another lesson I think we need to get clear on what the purpose of this lesson section is. Then, I can help edit if you want.

I assume that everybody encountering this lesson is expected to have completed the other lesson. I'd suggest we leave out the `row-n` section from this lesson entirely. And, instead, we should reframe the second section be contrasting the column look up with what `row-n` is doing.  Thoughts?

flannery-denny:
@schanzer In my memory, the starting point for all of this work was to address #2263. Excuse me if I'm missing something, but the original issue does not seem to have been addressed at all in this pull request. Probably worth jumping on the phone when you have a chance?

schanzer:
@flannery-denny see my comment on the PR itself. The original issue was a bad attempt at fixing a larger problem, and should be ignored

flannery-denny:
Tasks for @schanzer per our discussion.

1. Consider whether the Tables-Rows and Visualizations lesson should be split in two: 1 lesson for tables and another for visualizations.
2. Revisit the Lookups lesson. Likely discard of the first section. Definitely give the second section some love. Flesh out the lesson plan writing. Likely split the second section into sub sections.

schanzer:
@flannery-denny I've split the contracts lesson up into two, and removed some content from lookups. I'm afraid I can't get rid of the first section entirely, as it contains critical content that's used later in the lesson (and other lessons, too!)

Ready for your review.

lessons/Data-Science/lookups/langs/en-us/index.adoc
```diff
@@ -122,31 +90,88 @@ It's often better to name our Row definitions according to the _property we care
...
+- Turn to @printable-exercise{lookup-expressions.adoc} and notice that the code at the top uses the "age" of `old-row` to determine the size of the triangle.
+- Type the code into the Interactions Area and test it out.
+- With a partner, work through @printable-exercise{lookup-expressions.adoc}
+}
+
+@teacher{
+Have students share their most creative uses of lookups. Did anyone use three or more columns from a Row? Did anyone use more than one Row in the same expression?
 }
 
 === Synthesize
```
flannery-denny, line 161: @schanzer This is a pretty unorthodox Synthesize section in that it looks forward rather than backward. I'm not totally sure that it works. They don't actually know what a Table Function is yet and they're being asked to reason about them.

lessons/Data-Science/lookups/langs/en-us/pages/lookup-expressions.adoc
```diff
@@ -0,0 +1,60 @@
...
+<style>
+/* For the example prompt, keep the lines tight and centered */
+.tightAndCentered * { padding: 0 !important; text-align: center; }
+</style>
+++++
+
+Make sure you've opened your *saved copy* of the @starter-file{animals}, in which you've defined the sample Rows from @printable-exercise{defining-rows.adoc}, such as `lizard-row`, `old-row`, etc.
+
+[.tightAndCentered]
+The code below draws a solid, green triangle, using 5x the age of `old-row` as the size: +
```
flannery-denny, line 12: The description and the code don't match here. Please decide whether you want it to use the age or 5x the age.

PR #2393, opened 2025-02-07
Linear Overhaul

flannery-denny:
This pull request represents a significant rewrite of our linear modeling materials with

- new pages
- optional sections in lesson plans
- more interpretation stems and guidance around comparing models
- more solutions and screenshots

I expect it will need a read through from fresh eyes.


lessons/Algebra2/linear1-exploring-states/langs/en-us/pages/exploring-states.adoc
```diff
@@ -13,7 +23,11 @@ Then, click "Run" and type `states-table` into the Interactions Area on the righ
 |
 |===
 
-@n What code will produce a table showing the number of states in each region? @fitb{}{@ifsoln{@show{(code '(count states-table "region"))}}}
+
+== Digging into the Code
```
schanzer, line 27: This set of questions is a great way to keep `sort` and `first-n-row` fresh in students' minds. But is that actually important for Alg2? I'm not opposed to keeping these questions, but it's worth 5m of talking about whether they should go on an optional page instead

lessons/Algebra2/linear2-building-models/langs/en-us/index.adoc
```diff
@@ -115,15 +126,24 @@ Building on prior knowledge of linear functions, students learn to find the line
 This scatter plot appears to show a positive, linear relationship: +
 _States with higher percentages of college graduates tend to have higher median household incomes._
 }
-@ifnotslide{@center{@image{images/college-v-income.png}}}
+@ifnotslide{@center{@image{images/college-v-income.png, 450}}}
 
 @slidebreak
 
 @QandA{
 @Q{What do you notice about the @vocab{Form} of this scatter plot? What pattern do you see?}
```
schanzer, line 134: Now that the correlations lesson isn't required, I don't think we should use the vocabulary word `form` here. We also don't _need_ it! All we really want to ask is what shape the relationship looks like, right?

lessons/Algebra2/linear2-building-models/langs/en-us/pages/model-college-v-income-1-scaffolded.adoc
```diff
@@ -3,47 +3,50 @@
 <style>
 /* Shrink fill in the blanks */
 .studentAnswerMedium { min-width: 8em !important; }
+.fitbruby {padding-top: 1rem !important;}
 </style>
 ++++
 
 [.linkInstructions]
 Open the @starter-file{alg2-states-intro}.
```
schanzer, line 11: I don't think it makes sense to have two separate files for this sequence. Let's discuss if you feel strongly about it.

lessons/Algebra2/linear4-other-forms/langs/en-us/index.adoc
```diff
@@ -13,8 +13,7 @@
 
 | Lesson Goals
 | Students will be able to...
-
--
+@objectives
 
 | Student-facing Lesson Goals
 |
```
schanzer, line 25: Need to add these

lessons/Algebra2/linear4-other-forms/langs/en-us/index.adoc
```diff
@@ -37,36 +36,57 @@
 
 |===
 
-== Other Forms of Linear Models
-@objective{linear-forms}
+== Optional: Horizontal Shift in Linear Functions
```
schanzer, line 45: Why optional?

schanzer (commented):
Overall this looks ready!

There are a few things I'd like to discuss, however, which could probably be knocked out in a 15m zoom or phone call. Let me know when/if you're around tomorrow or Thursday?

lessons/Algebra2/linear1-exploring-states/langs/en-us/pages/exploring-states.adoc
flannery-denny: agreed to remove these questions about building tables.

instead - things like find a state that has a lot of veterans. if you pick two states that have lots of in

lessons/Algebra2/linear2-building-models/langs/en-us/pages/model-college-v-income-1-scaffolded.adoc
flannery-denny: predefine a model for `mi-ca` in the starter file instead of giving way `al-ak`

lessons/Algebra2/linear4-other-forms/langs/en-us/index.adoc
flannery-denny: move this to quadratic lesson where horizontal shift is introduced

flannery-denny: @schanzer I have started by literally cutting and pasting the section from linear to quadratic so that we can see it.

1.  I notice that horizontal shift is also optional in the quadratic lesson... so I want to check in before I do any further work.
2. Making the change real will require updating Desmos files for both lessons.

Issue #2271, opened 2024-11-03
Periodic Unit ToDos


flannery-denny:
**Periodics1**
- [x] On "Converting between angles": blank circle with space for degrees and radians. THEN ask kids to label the hours on the clock (maybe). Also ask kids to do major angles (0. 90. 180. 270. 360), not mid angles. Provide decimal _and_ irrational representations (when we give angles away in the workbook page and solutions offer both).
- [x] Right now unit clocks relies on a Desmos activity. It would be a win to rewrite it in Pyret, so students define their functions for the clock and use a reactor to see if they did it right. Obviously the reactor would be hidden in a context -- this might be a cool challenge to throw at Dorai or Ben! #2232 
- [x] Integrate #2231

**Periodics2**
- [x] Split into sections -- perhaps a separate section for some of the coefficients? Frequency is particularly tricky, so maybe it deserves its own section?
- [x] Remind students and teachers that we're focusing in on sinusoidal functions specifically, not periodic functions in general
- [x] Peaks and Valleys instead of peaks/troughs. Add desmos for vertical shift.

**Periodics3**
- [x] Break up the single section and workbook page. `modeling-recent-co2.adoc` is too crowded! The first part (1-6) should be a whole page, probably with some clarifying prose or questions to clear up common misunderstandings 
-- There were so many fill in the blanks and teachers had a hard time following directions (e.g. people have to come up with a horizontal value of the phase shift... we don't tell them to drop a line down from the point on the curve... find the date - drop a vertical to the x-axis)"
- [x] The second part (7-9) should be its own page. Maybe have students convert the model to sin, to give them more practice?

PR #2407, opened 2025-02-15
Periodics

schanzer:
This PR implements the improvements described in #2271 . 

**Periodics 1**
- [x] [reply to this comment](https://github.com/bootstrapworld/curriculum/pull/2407#discussion_r1985795817)
- [x] add unit circle to comparison of clock to degrees.
- [x] blue and green on the unit clock starter file are the opposite of blue and green on the image on the workbook page.
- [x] make question 3 on reasoning about unit clocks take less vertical space so that this page fits on a single page.
- [x] When I close the reactor window I see the following. It doesn't mean anything. We need to provide teachers some guidance on how to handle student questions about it.

Screenshot 2025-03-07 at 3 38 33 PM

- [x] The lesson title does not match the pattern used in other sequences --> Change to Exploring Ferris Wheel Data
- [x] fix lesson description
- [x] update lesson goals to reflect actual lesson content
- [x] Fix this image so that it's actually isosceles `images/1-30-triangle.png, 150}` on `unit-clock.adoc` (It wouldn't hurt to also improve the unit circle image on this page.
- [x] add @ifsoln to `unit-clock.adoc`
- calculations for 1:30 to question 3.
- guidance for the last two questions
- [x] It's confusing that 
Screenshot 2025-03-07 at 3 51 35 PM

- the solution image for the graph they're drawing uses blue and green curves while reactor uses red and blue curves. 
- students are not told what colors to use to draw the curves.
- and immediately following the image we then ask them to describe the red and blue curves (from the reactor)
- [x] The [reactor output](https://code.pyret.org/editor#share=1SpJmgBqgTQ88WZPSGMPbV9DpI4hcpVLh) would be improved by adding a 1 and -1 to the vertical axis on the left side of the window where the curves are being drawn.
Screenshot 2025-03-07 at 3 38 26 PM

- [x] I don't know what this lesson instruction is supposed to say, but what it says doesn't make sense

> In our clock animation we have 12 "slices", and drawing 4 evenly-spaced labels around the clock.

**Periodics 2**
- [x] [resolve how we surface ferris wheel data](https://github.com/bootstrapworld/curriculum/pull/2407#issuecomment-2721786250) - we're moving these questions and a the screenshot to an opt-printable page
- [x] confirm that it makes sense for `vocab-from-diagram.adoc` to include both sections rather than putting these diagrams on two separate pages.
- [x] make a solution version of `ferris-wheel-scatter-plot-w-curve.png` to include in the lesson plan and `vocab-from-diagram.adoc`

**Periodics 3**
- [x] confirm that both colors of dots on the scatterplot are visible and distinguishable when printed in black and white. If not, maybe change them to solid and hollow?
- [x] reread the lesson to confirm that it works post surgery

**Periodics 4**
- [x] Currently this lesson launches with work that has already been tackled in periodics3. Move that work to this lesson.
- [x] [improve interpretation stem](https://github.com/bootstrapworld/curriculum/pull/2407/files#r1992238668)  
- [x] [~provide scaffolding for frequency in the lesson plan~](https://github.com/bootstrapworld/curriculum/pull/2407/files#r1993908300)
- [scaffold comparing s-value for the same model on different data](https://github.com/bootstrapworld/curriculum/pull/2407/files#r1992244311) - we decided this should be an opt-printable, but there is room on `predicting-co2.adoc`
- [x] finish reading the lesson to make sure that it flows post surgery.

PERIODIC4 REWRITE PROPOSAL

- [x] Launch:
- perhaps make a sketch of each kind of model we've studied and describe its characteristics: L,Q, E, log, periodic
- then - n&w / what form of model would you choose for this scatter plot (co2-table): L, Q, E, log, periodic?
- [x] page 1: full dataset
- Let's see how well a linear model would have fit the data by making an lr-plot.
record s-model.
- Fit the periodic model to the full dataset. record s-model.
- practice interpreting both models?
- which is better. how much better?
- [x] page 2: modern table
- change modern-table image on workbook page to fit-model
- ask where the model fits the data best & worst
- how would the describe a model that would fit this dataset well?
- Table of s-values for recent table, modern table, co2 table
- Why doesn't it make sense to compare them?
(unclear whether the table should include range or let them come up with that idea. unclear whether we should give away s-values or ask them to find them all. )
- [x] page 3: combining models
- expand scaffolding for hybrid model.
- what kind of line does our periodic model wrap around?
- what is the equation of that line?
- where does it show up in our function definition?
- which line do we wish our model wrapped around?
- then separate the function into two separate functions - tell don't ask about how this works.
- then ask them to try to combine their models into a hybrid model
- fit model
- how does s compare to the lr-plot model we used for the full dataset
- [x] new starterfile that use function notation to build hybrid
perhaps gives `line` as horizontal line and they have to replace with optimal linear function?
perhaps gives `wave` in cosine instead of sine and they can swap?
- [ ] add something about k being horizontal line the wave wraps around to periodic2 when we're introducing the model.


lessons/Algebra2/periodic1-exploring-rides/langs/en-us/index.adoc
```diff
@@ -135,142 +140,169 @@ Students begin to reason about points on the circumference of a circle, with the
...
 
-==== Radians
+Would the shape of the curves change if we used 360 degrees instead of 12 hours?
+
+@lesson-instruction{
+In our clock animation we have 12 "slices", and drawing 4 evenly-spaced labels around the clock.
+
+- Return to the @starter-file{alg2-unit-clock}, change `num-slices` to 360, and click "Run". Did the curves change?
+- Take a minute to play with `num-slices` and `num-labels`, making sure that `num-labels` divides evenly into `num-slices` with _no remainder!_
+- Can you divide the clock into 70 slices? 92?
```
flannery-denny, line 230: @schanzer I'm unclear what you want people to say here. I can type num-slices 70 and num-labels 5 into the starter file. But is this evenly divided? the clock, yes. the dots on the graph, less clear. Do we want to ask a more concrete question the elicits 5 as a possible response? Definitely need to provide more guidance for teachers!
Screenshot 2025-03-07 at 4 25 45 PM

schanzer:
@flannery-denny I've completed the checkboxes above. For `vocab-from-diagram.adoc`, I actually think the best way to exercise students' understanding of vocabulary by having them extract answers instead of labeling a graph. If you disagree, let's talk about it -- but I've also made the solution image you requested

flannery-denny:
> @flannery-denny I've completed the checkboxes above. For `vocab-from-diagram.adoc`, I actually think the best way to exercise students' understanding of vocabulary by having them extract answers instead of labeling a graph. If you disagree, let's talk about it -- but I've also made the solution image you requested

@schanzer that's fine by me. confirming that means the todo is to move the pyret graph to its own page that has accompanying questions?

lessons/Algebra2/periodic3-fitting-models/langs/en-us/pages/modeling-recent-co2-2.adoc
```diff
@@ -40,31 +43,98 @@ table+.paragraph .autonum {padding-top: 0; }
...
+@n Use `fit-model` to fit your `periodic` model to the data in the `recent-table`. What S-value do you get? @fitb{}{@ifsoln{Answers will vary slightly.   ~1.2 }}
 
-@n *What does this model actually mean?* Fill in the blanks below, and read the completed model aloud with your partner.
+@n What does this model actually mean? Fill in the blanks below, and read the completed model aloud with your partner.
 
 @vspace{1ex}
 
 @indented{
 _Between the end of 2022 and 2023, the amount of @math{CO_2} in the air fluctuated between @fitbruby{6em}{@ifsoln{415.74}}{lowest} and @fitbruby{6em}{@ifsoln{424}}{highest} parts-per-million. This pattern appears to be @vocab{periodic}, with an amplitude of @fitbruby{6em}{@ifsoln{4.13}}{amplitude} rising and falling around a @vocab{midline} of @fitbruby{6em}{@ifsoln{419.87}}{midline}. With 1 year representing a full cycle, we expect this pattern to repeat each year for a frequency of @fitbruby{5em}{@ifsoln{@math{2pi}}}{frequency}._
 }
```
flannery-denny, line 87: @schanzer 

- [ ] we have to do a better job with helping students to make sense of frequency than this interpretation stem does. I would fix it myself, but I don't actually know what it means off the top of my head.
- [ ] The interpretation stem should also engage students with the S-value. On other pages we say 

> The error in the model is described by an @vocab{S-value} of about @fitbruby{1em}{@ifsoln{826}}{S}@fitbruby{8em}{@ifsoln{positive cases}}{units}. @hspace{1em} I
>  @fitbruby{18em}{@ifsoln{disagree}}{strongly agree, agree, disagree, strongly disagree} that this model is a good fit considering that @fitbruby{8em}{@ifsoln{positive cases}}{y-variable} in this dataset range from
>  @fitbruby{8em}{@ifsoln{1092}}{lowest y-value} to @fitbruby{8em}{@ifsoln{6781}}{highest y-value}.}
>

lessons/Algebra2/periodic3-fitting-models/langs/en-us/pages/modeling-recent-co2-2.adoc
```diff
@@ -40,31 +43,98 @@ table+.paragraph .autonum {padding-top: 0; }
...
+
+@vspace{1ex}
+
+[cols="10a,1,10a", grid="none", frame="none"]
+|===
+| @n Sketch your model on the scatter plot below.
+
+@ifnotsoln{@center{@image{../images/modern-co2-scatter.png, 300}}}
+@ifsoln{@center{@image{../images/modern-co2-model-soln.png, 300}}}
+
```
flannery-denny, line 103: @schanzer This seems like an opportune moment to have students use percent change to compare s-values

Here's how we do that in fitting models

> @n How much
> @fitbruby{7em}{@ifsoln{more}}{more / less }
> error do we expect in predictions made with Jo's model than predictions made with Cy's model?
> 
> @vspace{1ex}
> 
> @scrub{
> A note for future authors:
> - &#8192; is used below to add horizontal space.
> - \mbox{} is used to make the words in the fraction use normal font.
> }
> 
> @indented{
> Percent Change &#8192; = &#8192;
> @math{
> \frac{&#8192; \mbox{Difference} &#8192; \mbox{between} &#8192; \mbox{the} &#8192; \mbox{S-values}&#8192;}
> {\mbox{S-value} &#8192; \mbox{for} &#8192; \mbox{Cy's} &#8192; \mbox{model}}
> \times 100 &#8192; = &#8192; }
> @ifnotsoln{@math{\frac{\qquad}{\qquad}}}
> @ifsoln{@math{\frac{4.3}{5.9} = .73  &#8192; &rarr; &#8192;  73 \%}}
> }
> 
> We expect predictions made with Jo's model to have
> @fitb{4em}{@ifsoln{73}} percent
> @fitbruby{7em}{@ifsoln{more}}{ more / less }
> error than predictions made with Cy's model!

schanzer:
> > @flannery-denny I've completed the checkboxes above. For `vocab-from-diagram.adoc`, I actually think the best way to exercise students' understanding of vocabulary by having them extract answers instead of labeling a graph. If you disagree, let's talk about it -- but I've also made the solution image you requested
> 
> @schanzer that's fine by me. confirming that means the todo is to move the pyret graph to its own page that has accompanying questions?

No - I don't think moving the graph is actually helpful. There's very little mental activity involved in comparing the reference image to the pyret graph, and asking them to label the parts. I think we should skip it entirely, in favor of the new activity I've replaced it with

flannery-denny:
> > > @flannery-denny I've completed the checkboxes above. For `vocab-from-diagram.adoc`, I actually think the best way to exercise students' understanding of vocabulary by having them extract answers instead of labeling a graph. If you disagree, let's talk about it -- but I've also made the solution image you requested
> > 
> > 
> > @schanzer that's fine by me. confirming that means the todo is to move the pyret graph to its own page that has accompanying questions?
> 
> No - I don't think moving the graph is actually helpful. There's very little mental activity involved in comparing the reference image to the pyret graph, and asking them to label the parts. I think we should skip it entirely, in favor of the new activity I've replaced it with

@schanzer the page is fine, but if it doesn't include the image it needs to direct students to the dataset and at the point that this workbook page appears in the lesson plan, students do not yet have access to the graph of the ferris wheel data. You seem to have strong opinions about this, so, before I fix the lesson plan I need to know whether you want them to look at the data sooner than they currently do or you want them to circle back to the second section of this page later than they currently do. happy to zoom about this if that's easier.

lessons/Algebra2/periodic3-fitting-models/langs/en-us/index.adoc
```diff
@@ -139,14 +146,56 @@ The `recent-table` includes just the rows from trough-to-trough for the years 20
 @A{One year (this makes sense, since the seasonal cycle repeats every year!)}
 @Q{What did you get for @vocab{frequency} @math{b}?}
 @A{@math{2\pi}, because the period is 1 year and @math{{2\pi \over 1} = 2\pi}.}
+@Q{What does this frequency tell us?}
+@A{}
```
flannery-denny, line 150: @schanzer here's another spot of helping students make sense of frequency.

lessons/Algebra2/periodic1-exploring-rides/langs/en-us/index.adoc
flannery-denny: Someone decided that there are 360 degrees in a circle.
Someone decided that there are 2pi radians in a circle. 
What number of slices would you use for a circle if you were in charge of making a new labeling system? what would the num-labels be - sketch the clock as it would appear in this pyret file.

lessons/Algebra2/periodic3-fitting-models/langs/en-us/pages/modeling-recent-co2-2.adoc
flannery-denny: tackled this

lessons/Algebra2/periodic3-fitting-models/langs/en-us/pages/modeling-recent-co2-2.adoc
flannery-denny: decided to make an opt-printable for this.

schanzer:
@flannery-denny Since I’m not added as a reviewer on the PR itself, I’m writing my review of periodics4 as a comment here. Overall, I think this is a _fantastic_ rewrite! Comments on individual files below:

**"Choosing the Best Model for the Data”** 
	- For the sketches, why bother having them draw axes at all? Let’s just give them empty "graphing paper squares” to get them started
	- For the key characteristics, why not give them lines on which to write?
	- Why not give away the sketch and key characteristics for linear models? It makes it abundantly clear what we want for the rest, and makes the page less time-consuming. By this point, asking them to draw a linear model and describe its shape feels more remedial than review
	- I’m concerned about the spacing on the page: the graphs don’t need to be large for kids to draw a sketch, and every other part of the page needs more vertical space (and blank lines!). The last question on the page has very little room for an answer, and no blank lines at all. I suggest shrinking the empty graphs and the scatter plot at the bottom, to make more room for student responses.
	- In the lesson plan, we talk about the co2-table from the wbpage, but never actually have students make it before the discussion. Why not have them make it, talk amongst themselves, and then do this workbook page? Showing them such a different plot with zero context seems counter-productive to me.

**Modeling Modern Carbon Dioxide Levels**
	- Why ask them to write the linear model in both function and pyret notation? At this point, going straight to pyret from lr-plot should be pretty easy
	- “update” is too vague. If we want students to change some code, let’s use the word change
	- It would be nice if students had to sketch all three graphs on the scatterplot: the linear one, their periodic one, and their imaginary “best one”. Drawing out the “best one” before describing it would be extremely helpful!
	- In fact, I could see getting rid of the description altogether, and giving students more space to draw on the scatter plot. The teacher could solicit verbal descriptions from students, asking them to describe what they drew.

**Building a Hybrid Model (I think this page needs the most work)**
	- "What kind of line does the wave wrap around?” are you just asking them to say “linear”? or “horizontal”? I’m not sure if this is adding much, when we really just want them to identify the equation of the midline. 
	- Would it make sense to rename “trend-line” to “midline”?
	- I’m finding it hard to follow the order of operations on this page. We start by having them work with their sin-based periodic model and ask a bunch of questions, but then we switch to the cosine-based model and ask a different set of questions. After that we return to the sine-based one. In the process they have to look at six different function definitions.
		-> Could we instead just start with the cosine-based one? All of question 1 still applies, since it has the same midline.
		-> Then it’s a direct link to making sense of the wave, trend, and hybrid definitions.
		-> Rather than giving students the “wave” function for their own periodic model, I think we should ask them to define it themselves, calling it “wave-sin” to match “wave-cos”. If they can’t code it themselves, then they aren’t understanding what’s happening enough to make sense of the code if we give it to them anyway
	- hybrid-cos is a confusing name, because there’s nothing “hybrid” about it!
	- “update” is too vague. If we want students to change some code, let’s use the word change
	- "What is the S-value of hybrid-modern?” this doesn’t make sense. A model doesn’t have an S-value! Only a model fit to a particular dataset can have an S-value.

**More Models**
	- If we’re using a function called hybrid-modern to fit modern data, shouldn’t we call the function that fits all the data hybrid-all?

--------------------------------
I pushed a significant change to `hybrid.adoc`, which I think moves us closer to where we want to be. Otherwise, my commit is mostly wording changes.

PR #2444, opened 2025-03-14
ai-intro + ai-training are ready for review

retabak:
RT: ai-training is being split into two lessons; the second lesson is still being improved!

FD: Heads up - seeing these two-page warnings when I make book.
Screenshot 2025-03-14 at 11 52 29 AM

- [x] For people who don't live in your head, it would be helpful if you could update the `lesson-order.txt` file for the pathway as it changes so that we can get a quick overview of the bigger picture plan before digging into the nitty gritty of an individual lesson.
- [x] please please please save all drafts of starter files in the shared google drive so that you are not the gatekeeper to information and problem solving when things don't work. Our systems are designed for collaboration and so that others can pick up the pieces wherever they are left off and I keep getting blocked by the fact that you aren't using them.

**ai-intro**
- [x] update `learningObjectives/ai.json`
- [x] suggest adding a simple handout or google form to the second section to keep students focused on the task (perhaps teachers would also want to collect it?)
- name of media chosen.
- checklist of statements for them to select one from
- space to write about how the chosen media supports the idea

**training**
- [x] I can't actually test out your `a slightly less primitive plagiarism detector` page because the Plagiarism Detection Starter File does not have the details needed to follow the directions and you also haven't provided solutions that would let me imagine what is supposed to be happening without actually using the starter file.
- [x] from what I can see you have not improved the plagiarism starter file per our discussion.
- [x] are `primitive plagiarism detector` and `a slightly less primitive plagiarism detector` supposed to use the same starter file? 
- If so, update the directions in the lesson plan to make that clear.
- If not, update the links to take us to a different starter file (based on how much doesn't work It feels pretty likely that you have a secret second starter file somewhere that you haven't linked to the lesson)


lessons/AI/ai-intro/langs/en-us/index.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+But what *counts* as AI?
+
+@lesson-instruction{
+- You are going to complete a @online-exercise{https://docs.google.com/forms/u/1/d/1w3oOSXTuMzCLLRDl7zh4YH_ddry2vUIGSSYOpqPWh-U/copy, "survey"} where you will decide whether seven different technologies represent AI or not.
+- Not sure if it's AI or not? You can choose "I have no idea!" No need to overthink.
+}
+
+@teacher{
+This survey is designed to get students thinking about what constitutes AI. Once students have completed the survey, debrief using the questions and text below. The goal is to help students recognize that the best answer, in this case, is almost always "I have no idea!"
```
flannery-denny, line 53: Please provide some guidance about why the answer should be "I have no idea!"

PR #2473, opened 2025-04-07
Card sorts

retabak:
Card sort solutions!

I also noticed that the prerequisites for the gold standard lessons were not accurate, so attempted to update those on this branch.


flannery-denny:
@retabak I think this is about ready to squash and merge. My one suggestion is that you use the `@teacher{}` objective on the solution pages for the notes you're writing that are sort of like asides. In the workbook it will print **Teachers:** preceding whatever is inside the directive.

PR #2509, opened 2025-04-27
Incorporate live surveys into DS lessons where we introduce new displays

flannery-denny:


flannery-denny:
- bar and pie charts
- dot plots
- dot plots to histograms
- scatter plots
- measures of center
- box plots
- linear regression
- css work around opt-blocks

PR #2751, opened 2026-03-18
Ee2

flannery-denny:
- Agree on True or False and Which one isn't Equivalent formatting
- Move Programming to start of lessons
- More work on OoO

### DIFF
--- a/lessons/Expressions-and-Equations/equivalence/langs/en-us/pages/true-or-false-images.adoc
+++ b/lessons/Expressions-and-Equations/equivalence/langs/en-us/pages/true-or-false-images.adoc
@@ -76,7 +76,7 @@ For each question below, review the collection of expressions. Place an *E* in f
 ! @ifsoln{E}  ! @show{(code '(rotate 90 (rectangle 48 72 "solid" "blue")))}
 ! @ifsoln{E}  ! @show{(code '(scale 8 (rectangle 9 6 "solid" "blue")))}
 !===
-| @ifsoln{When we rotate a rectangle 90 degrees it places the long side where the short side was, so a 48 x 72 rectangle becomes a 72 x 48 rectangle. `scale` tell use to multiply the base and height times the scale factor: @math{8 \times 9 = 72} and @math{8 \ times 6 = 48}.}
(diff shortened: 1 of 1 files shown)


lessons/Algebra/order-of-operations/langs/en-us/index.adoc
schanzer: Overall I'm very happy with these changed! I made some small edits/additions, but I feel good about where it's at. 

Now that it's in the E+E pathway, I wonder: are we introducing circles twice when we don't need to, since we first do it in the Translating lesson? Or does it come across as a useful refresher to see it here?

lessons/Expressions-and-Equations/distributive-property/langs/en-us/index.adoc
```diff
@@ -28,10 +28,100 @@
...
+@A{Possible Wonderings: 
+  * Are these equivalent? 
+  * What do they evaluate to? 
+  * Why is there one `scale` in the first Circle of Evaluation, but two 9s in the second Circle?
+}
+}
+
+@slidebreak{Launch}
+
+Sometimes instead of combining two objects before transforming them, we can transform each object and then combine them. Let's take a look at what's happening inside each of these Circles of Evaluation.
```
schanzer: This is a nice place to talk about _efficiency_. Suppose, for example, that rotation was fast and easy, but scaling was very expensive? (In a world where GPUs are literally firing up AI just to fill in missing pixels when a frame of a videogame is scaled, this happens to be true!)

The first COE (scale is distributed) would actually run quite a bit faster than the second.

schanzer (commented):
Overall I see a LOT of big improvements. The way you've started the lessons with an image-based programming exploration is really fantastic. I know it's a WIP, but here are some meta-points that apply in several places:

- Some of the opening explorations use wbpages that used to live at the end of the lesson, so they're written as if students already know what property X is. This language could use some cleaning up in places.
- You're using variables a lot earlier than the lesson itself appears in the pathway. I don't think it's bad to use them earlier (and I definitely see the added value!), but I suspect this means the variable lesson itself will need to either be moved, substantially rewritten, or removed altogether.
- The programming intros aren't as uniform as they could be. I LOVE LOVE LOVE the way you motivated the commutative one as a programming need! I think you should do that everywhere. (Also, the programming activity in absolute-value-and-opposite is at the end. Will you move it to the beginning?)

PR #2796, opened 2026-06-14
Js cleanup

schanzer:
bootstraplesson.js has grown out of hand. 

This PR splits it into multiple files, which is extra nice since 95% of our generated files only need 5% of the JS. :P


ds26gte:
From the build side (i.e., changes to Makefile*, do-postproc.lua), everything looks OK.  I'll defer to you about the actual JS content split (i.e., whether any JS code got excluded/included that shouldn't).

PR #2798, opened 2026-06-15
Stale cache fix

schanzer:
Repeated, unforced `make`s have surfaced what appears to be a bug in the way Phase 1 hands off the state of the distribution directory to Phase 2. Claude has a proposed diagnosis and fix, but something about it doesn't feel right. @ds26gte, you know the system best, so I'd like you to see if you agree with the diagnosis first and the fix second

 (Read Claude's commit message to see what it thinks the issue is)

### DIFF
--- a/lessons/AI/decision-trees-building/langs/en-us/index.adoc
+++ b/lessons/AI/decision-trees-building/langs/en-us/index.adoc
@@ -499,15 +499,8 @@ Use one of our new tools to answer the following questions about our `tail-class
 }
 
 @QandA{
-@Q{How do we choose a root node that will make the best possible split?}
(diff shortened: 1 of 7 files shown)


ds26gte:
Temp files ending in .kp are supposed to be persistent across runs (kp was my shorthand file-extension for "keep"), but I guess unusual circumstances are preventing them from being made sometimes? If this is indeed happening, the changes look ok and will not harm.

I notice that the check is being conducted only for 
- workbookpage_adocs (line 91ff)
- lessonplan_adocs (l. 136ff)

but not for

- pathwayindependent_adocs (l. 76ff)
- pathwayresource_adocs (l. 186ff)
- pathwaynarrative_adocs (l. 239ff)

PR #2799, opened 2026-06-15
Fix empty kp hang

schanzer:
The actual symptom of the stale cache bug: 

- at times, `massage-course` could leave an empty cache file
- since the file exists, the lua `sread` calls in Phase 2 try to parse it
- but it's empty, with no EOF tag -- so lua runs forever

The three commits in this PR adds a `nil` check to the `sread` calls, make sure that `merge-pathway-tocs` can throw an error and keep going rather than choking, and makes `massage-course` atomic so that it never generates an empty file

### DIFF
--- a/lib/maker/make-pathway-tocs.lua
+++ b/lib/maker/make-pathway-tocs.lua
@@ -23,6 +23,13 @@ for _,course in ipairs(all_courses) do
   local this_course_lessons_file = course_dir .. '.cached/.workbook-lessons.rkt.kp'
   if not file_exists_p(this_course_lessons_file) then goto continue end
   local lesson_units = sread_file(this_course_lessons_file)
+  -- An empty or malformed .kp makes sread_file return a non-table (false on
+  -- empty/exhausted input). Skip the course rather than indexing into it.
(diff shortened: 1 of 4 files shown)


ds26gte:
As mentioned in private slack, an alternate fix -- a simple rewrite of `make-workbook-lessons-list.lua` -- in branch `fix-empty-kp-hang-within-lua` appears to be enough.

schanzer:
@ds26gte I appreciate the poke! I'm testing your fix locally today on my local machine, and if the problem doesn't re-appear I'll merge

schanzer:
@ds26gte Confirming that this fixes one issue, but not all of them.

Your fix eliminates one of the sources of a 0-byte files (pathways with no lesson-order.txt files), but not all of them (a halted `make`). As you pointed out, this branch is _complimentary_ in that it addresses the infinite loop by modifying `sread_file` to watch out for 0-byte files in the first place. But it also fixes the _crash_ by adding a guard to `make-pathway-tocs` _and_ makes `massage-course` atomic by using the temp file. 

I've cherry-picked your change into this branch, and will merge it to master after some testing.

